Issue #224, opened 2026-06-27
Feature parity: explicit preserve-vs-rebuild decision per interactive region (verbatim JS only on verbatim markup)


chubes4:
## Problem
The transform rewrites the DOM (e.g. a JS hamburger toggle -> core/button or dropped, #mobile-nav deduped), but the original JS targets the source selectors (.nav-toggle / #mobile-nav). So when that JS is enqueued it loads and **silently no-ops** — the markup it needs no longer exists. Behavior is "preserved" on paper but inert.

## Ask
Make a per-interactive-region decision explicit:
- **Rebuild**: emit native interactivity (core blocks / Automattic products / Interactivity API) and DROP the original JS for that region. Robust, DOM-independent.
- **Preserve**: keep the region's markup **verbatim** as a `core/html` runtime island AND carry its JS (via the companion plugin, scoped to the island). Works because the markup is untouched, so the original selectors still match.

**Hard rule:** verbatim JS only on verbatim markup. Never enqueue original JS to drive transformed blocks. The recognizer/island machinery should tag each region preserve-vs-rebuild so downstream (JS materialization, diagnostics) can rely on it.

chubes4:
Implementation plan for the typed carousel rebuild:

This is a preserve-vs-rebuild application of the existing runtime-island architecture. For a high-confidence, bounded carousel, Blocks Engine will rebuild into a generic typed companion declaration rather than preserve source script execution and multiple raw HTML islands. The declaration retains ordered content, presentation semantics, control/index/wrap behavior, labels, per-instance scope, and source selector/style lineage.

The initial implementation is intentionally bounded to translated testimonials from fixture 87, while fixture 79 proves separate active-class recognition. It will add negative rail/non-carousel tests and assert that only rebuilt-region scripts are suppressed. SSI #734 owns editor-capable block materialization and the WordPress Interactivity API runtime.

AI assistance: gpt-5.6-terra via OpenCode was used to inspect the architecture and prepare this implementation plan. Chris Huber remains responsible for every line.

chubes4:
Investigation evidence: fixture 87 has one external `js/site.js` containing both reveal behavior and the translated testimonial handler. Current artifact compilation packages that whole source into `preserved_js` through an unrelated preserved form runtime island. The companion payload contract records only script content/selector; it has no source-script-to-rebuilt-region ownership mapping and no safe JavaScript module/slice representation.

[...]

PR #273, opened 2026-06-28
Populate companion preserved_js from runtime islands (+ fix silent JS loss in generated blocks)

chubes4:
## What
Populates `CompanionPluginPayload.preserved_js[]` from runtime-island data so behavior JS embedded in generated custom blocks is carried into the SSI companion plugin (scoped per block), instead of being lost. The consumer already merged (SSI #496); this builds the producer half (#488).

## Notable: fixes a silent JS-loss bug
While building the island→block association (which did **not** exist — generated blocks and runtime islands came from disjoint elements, and `sanitizeHtmlString` stripped `<script>` bodies from generated content), I found inline behavior scripts inside generated custom blocks were being **silently dropped**. The fix mints the association at generation time.

## Change (`php-transformer`)
- `FallbackEmitter::maybeGenerateCustomBlock()` now scans the generated subtree for runtime `<script>`s and records each as a `script` island tagged with the owning block's FQN via a generic `owner_block` field (shared `scriptIslandMetadata()` helper).
- `RuntimeIslandPackageBuilder` carries `owner_block` through (stays product-neutral).
- `CompanionPluginPayload::preservedJs()` projects islands → `preserved_js`, gated by #224: only `disposition=preserve` + `js_handling=preserve_verbatim`, skipping telemetry/`droppable`. Per entry: `content` (verbatim JS), `handle`, `src`. Islands with a matching `owner_block` are **scoped** (`block=ssi-<slug>/<name>`); preserve-worthy islands with no sound owner go to **`preserved_js_deferred[]`** with a `reason` — never scoped by guess, never silently dropped.

## Scope (honest)
- **Built end-to-end:** scripts living inside a generated custom block → scoped, enqueued only when that block renders.
- **Deferred (declared):** free-standing islands (standalone scripts, canvas/form) — no sound generated-block owner exists yet (`reason: no_generated_block_owner`), the #488 follow-up.
- External unmaterialized scripts: skipped (consumer requires `content`). Materialized external: carried.

## Verification
- `composer test` + `composer parity` → **144 fixtures green**; new contract coverage asserts the populated scoped entry, the deferral path, and that telemetry JS never reaches the payload. Parity unaffected (only adds source-report data; serialized output unchanged).

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Code (Claude Opus 4.8, 1M context)
- **Used for:** Verification, the association design, implementation, and tests under human review.

### DIFF
--- a/php-transformer/src/ArtifactCompiler/ArtifactCompiler.php
+++ b/php-transformer/src/ArtifactCompiler/ArtifactCompiler.php
@@ -77,14 +77,20 @@ final class ArtifactCompiler
         );
         $sourceReports['compiled_site'] = $this->compiledSiteReport($normalized, $entryPath, $documents['documents'], $assets, $blockTypes, $serializedBlocks);
         $sourceReports['materialization_plan'] = ( new MaterializationPlanBuilder() )->fromCompiledSite($sourceReports['compiled_site']);
-        $companionPluginPayload = $companionPluginPayloadBuilder->fromBlockTypes($blockTypes, $normalized['files'], $artifact, $entryBlocks['generated_blocks']);
+        // Build the generic runtime-island package first so the product-named
+        // companion-plugin payload can map preserved island JS into its
+        // preserved_js slot (issue #488). The package is the engine-neutral feed;
+        // CompanionPluginPayload owns the SSI-named projection.
+        $runtimeIslandPackage = array() !== $entryBlocks['runtime_islands']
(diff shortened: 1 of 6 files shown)


chubes4:
Parking as superseded. End-to-end verification showed the companion-plugin / preserved-JS path is the wrong mechanism for interactivity: custom-block generation is structurally inert across all 72 website fixtures (0 generated — the generation gate at the html_unsupported_element tail is unreachable), and generated blocks strip <script>/on* via wp_kses_post anyway. The real interactive behavior is delivered deterministically by native conversion (core/navigation responsive overlay, core/details + core/accordion). Closing; branch retained for reference.

PR #555, opened 2026-07-06
Reduce Figma HTML flow scaffolding

chubes4:
## Summary
- Adds a safe content-only flow scaffold path for Figma HTML containers whose children are explicitly absolute but classified as layout-intent flow.
- Emits layout-intent attributes/styles for those safe normal-flow containers while avoiding nodes that still position absolutely.
- Adds a contract fixture proving absolute text rows in a content section enter normal flow without left/top offsets.

## Verification
- php figma-transformer/tests/contract/run.php
- FSE fixture regenerated before/after from latest origin/trunk using figma-transformer/scripts/figma-fixture-matrix.php with the local FSE .fig fixture.

## FSE Artifact Counts
- Before raw generated HTML/CSS count: position:absolute=119, pixel/calc offset declarations=580.
- After raw generated HTML/CSS count: position:absolute=121, pixel/calc offset declarations=584.
- Readiness summary remained: large_css_offset_count=0, large_absolute_offset_count=0.

## Blockers / Notes
- This PR proves the safe synthetic content-scaffold path, but the current FSE fixture did not improve. It regressed by two nested absolute rules, so this should be reviewed as an incremental guarded primitive rather than the final FSE reduction.
- I could not locate local Fisiostetic or TT5 .fig fixtures in /Users/chubes/Developer or this Studio workspace, so those sanity runs were not performed.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** openai/gpt-5.5 via OpenCode
- **Used for:** Drafted and verified the scoped Figma HTML transformer change, generated FSE artifacts, measured counts, and prepared this PR.

### DIFF
--- a/figma-transformer/src/Html/StaticHtmlEmitter.php
+++ b/figma-transformer/src/Html/StaticHtmlEmitter.php
@@ -1022,7 +1022,7 @@ final class StaticHtmlEmitter
         if ( null !== $semanticRole ) {
             $attributes .= ' data-figma-semantic-role="' . $this->sanitizeAttribute($semanticRole) . '"';
         }
-        if ( $this->layoutIntentShouldEmitClass($layoutIntent) ) {
+        if ( is_array($layoutIntent) && ($this->layoutIntentShouldEmitClass($layoutIntent) || ($this->freeformContainerShouldUseFlow($node) && ! $this->nodeWillPositionAbsolute($node, $parentNode))) ) {
             $attributes .= $this->layoutIntentAttributes($layoutIntent);
         }
         $anchorId = $this->headingAnchorId($node, $tag);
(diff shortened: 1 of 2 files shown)


chubes4:
Triage result: recommend **not merging / closing this PR as-is**.

Evidence from current PR head (`fb69fb00`, branch `fix/figma-flow-scaffolding-parity`):

- Contract tests pass: `composer test` in `figma-transformer`.
- Fisiostetic fixture is unchanged versus baseline: `position:absolute=25`, offsets=`156`, flow displays=`75` before and after.
- FSE Pilot fixture regresses the target metric: `position:absolute=119 -> 121`, offsets=`732 -> 736`, flow displays unchanged at `269`.
- The two added absolute rules are:
  - `.figma-node-11320-3036-4178-6475-heading-with-separator` now has `position:absolute;left:0px;top:0px`.
  - `.figma-node-11320-3036-4166-4896-frame-16` now has `position:absolute;left:0px;top:80px`.

Root cause: the heuristic broadens flow-scaffold/layout-intent handling, but in the real FSE fixture it turns already-flowing Featured Posts children into positioned layers instead of reducing positioned scaffolding. The added synthetic contract only proves a narrow fixture shape and does not cover the real regression.

I did not find a small safe direct fix that makes this PR improve FSE/Fisiostetic/contract metrics without risking layout quality. This should either be closed or replaced with a narrower PR backed by the FSE Featured Posts case as a regression fixture.

Issue #624, opened 2026-07-18
Preserve CSS-significant inline and media geometry during block conversion


chubes4:
## Problem

The post-#623 `15-saas` browser matrix still fails exact visual parity despite producing 265/265 editor-valid native blocks and zero `core/html` blocks. The aligned mismatch improved from 36.74% to 24.89%, but the remaining drift is measurable transformer loss rather than WordPress outer-wrapper noise.

This blocks Automattic/static-site-importer#489.

## Root causes

1. Selector-addressable inline `<span>` leaves are lowered to paragraph/mark structures. In the hero this changes an 18.14px inline label into a 28.05px paragraph and shifts the section by 9.91px. In `.lm-meta`, three independently styled pill spans collapse into one paragraph with mark children, losing sibling box geometry and author selector semantics.
2. Materialized SVG used dimensions do not preserve the source aspect ratio. At the same 574px width, the product SVG changes from 351.80px to 380px high.
3. CSS-significant structural group classes are dropped. The product right column loses its author identity and the grid row becomes 39.61px shorter, accumulating downstream position drift.
4. Lower impact: generated button wrappers are not layout-neutral when source styles target the anchor, changing the header CTA from 37.80px to 54.70px high.

## Expected generic behavior

- Preserve selector-addressable inline semantic leaves, sibling boundaries, classes, and attributes when conversion to paragraph/mark changes authored CSS semantics.
- Preserve resolved SVG intrinsic/used aspect ratio in generated image block metadata and rendering.
- Preserve author classes on structural groups when referenced by authored CSS or required for descendant/child selector compatibility.
- Keep generated block wrappers layout-neutral when source geometry belongs to the represented leaf.

## Acceptance criteria

1. Add deterministic transformer coverage for inline sibling pills, inline label line-height, SVG used dimensions, CSS-significant structural classes, and anchor-styled button wrappers.
2. Preserve native Gutenberg output and editor validity; do not introduce fixture-specific selectors or `core/html` fallbacks.
3. Rerun the single-fixture SSI matrix against `15-saas` with exact visual gating.
4. Report source/candidate dimensions and mismatch ratio; promotion remains blocked unless exact 1:1 parity and human screenshot approval are achieved.

## Reproduction

1. Check out Blocks Engine with #621 and #623.
2. From a fresh Static Site Importer checkout, run its documented fixture-matrix wrapper against Blocks Engine with the authored `saas` tag and a Lab runner.
[...]

PR #629, opened 2026-07-19
Project root child selector semantics

chubes4:
## Summary
- project selectors such as `body > *` onto stable source-element markers
- preserve authored root-child stacking contexts when WordPress inserts outer document wrappers
- skip non-rendering document metadata while allocating projection markers

Follow-up to #628 and #624.

## How to test
1. Check out this branch from a fresh clone.
2. Run `cd php-transformer && composer test`.
3. Confirm all canonical contracts, unit tests, 244 parity fixtures, and package install proof pass.
4. Inspect `php-transformer/tests/unit/author-selector-semantics.php` and confirm two source sections receive distinct semantic markers for `body > * { position: relative; z-index: 1; }`.

## Compatibility
- No public API or extension path changes.
- Existing selectors continue to apply to non-projected document children.
- The change preserves source semantics that would otherwise be broken only by generated WordPress wrapper insertion.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenAI GPT-5.6 Sol via OpenCode
- **Used for:** Visual attribution, root-cause analysis, implementation, regression coverage, and verification in collaboration with Chris Huber.

### DIFF
--- a/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
+++ b/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
@@ -1286,6 +1286,15 @@ final class HtmlTransformer
             return null;
         }
 
+        // Handle a safe SVG at a phrasing-to-block boundary before generic
+        // preservation rules see the SVG as an unsupported document fragment.
+        if ( 'svg' === $tagName && $this->svgNeedsPhrasingHost($element) ) {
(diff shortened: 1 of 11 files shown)


chubes4:
Runtime visual evidence falsified this approach: applying root-child stacking markers increased aligned mismatch from 18.16% to 20.43% and raw mismatch from 18.43% to 21.75%, while editor validity remained green. Reverted rather than merging a regression.

PR #728, opened 2026-07-31
Emit core/cover for background-image hero containers

borkweb:
## What

Adds `core/cover` emission to the php-transformer HtmlToBlocks pipeline. Flow containers (`div|section|article`) whose CSS background is a real image hero — background-image URL plus overlaid text content — now convert to `core/cover` instead of `core/group`, with exact save()-shape markup matching vendored `@wordpress/block-library` 10.0.0.

## How

- **`CoverStyleResolver`** (new): quote-masked CSS parsing for background URL (last-declaration-wins cascade), uniform-gradient dim ratio derivation, min-height (`px|vh|rem|dvh|svh|lvh`), focal point from `background-position`, hero size gate, and repeating-background detection. Declarations are memoized with a 64KB cap.
- **`CoverPattern`** (new): recognizer wired into the flow-container dispatch ladder (after details disclosure, before columns). Gates: background URL present → hero size gate (background-size cover OR min-height ≥ 200px/30vh/12.5rem) → not repeating → not a columns layout → children convert to text-bearing blocks → not a nav shell → not a multi-layer background. Non-matching containers keep today's `core/group` path byte-identical.
- **`BlockFactory::coverHtml`**: hand-rolled cover save markup per vendored `src/cover/save.js` — `` before the overlay `<span>`, span immediately before `wp-block-cover__inner-container` (CSS sibling selector depends on this), `dimRatioToClass` omission at 50, px `minHeightUnit` elision.
- **Gradient overlays**: uniform 2-stop gradients become `dimRatio`; design (non-uniform) gradients ride `customGradient` on the overlay span at `dimRatio: 100` — core's CSS maps dim ratio to span *opacity*, so a design gradient at dim 0 renders invisible. Mixed uniform+design stacks keep all layers in `customGradient` in source order.
- **Validation & diagnostics**: `core/cover` registered in `CanonicalSaveShapeValidator` TARGET_BLOCKS and `GeneratedGutenbergClassPolicy`; new `cover_gate_rejection` corpus detector (INFO) records background-image-with-content containers that fail the gates, so heuristic under-fire is measurable.

## Testing

- New unit suites for `CoverStyleResolver` and `CoverPattern`; corpus-detector cases extended.
- Six new contract cases: exact-equality hero golden, dim/no-double-paint, gate-rejected byte-identical fallback, cover-containing-columns, empty-container regression, style ordering.
- Parity goldens regenerated where heroes now convert to covers; multi-layer-background golden verified byte-identical (correctly rejected).
- Visual QA: real source/converted HTML pairs rendered in Chrome with the generated CSS and vendored `build-style/cover/style.css` — plain, dim-overlay, design-gradient, and mixed-stack heroes all paint identically to source. This caught (and drove the fix for) the invisible design-gradient defect.
- Corpus scan over fixtures/websites: zero covers emitted, verified correct — the only `url()` background rules present are repeating textures and a body background, all properly rejected.

Full `composer test` green.

### DIFF
--- a/php-transformer/composer.json
+++ b/php-transformer/composer.json
@@ -62,6 +62,7 @@
       "php tests/unit/css-value-splitter.php",
       "php tests/unit/background-image-extractor.php",
       "php tests/unit/cover-block-factory.php",
+      "php tests/unit/cover-pattern.php",
       "php tests/unit/cover-style-resolver.php",
       "php tests/unit/css-stylesheet-transformer.php",
       "php tests/unit/css-selector-matcher.php",
--- a/php-transformer/composer.json
+++ b/php-transformer/composer.json
@@ -61,6 +61,7 @@
       "php tests/unit/description-list-block.php",
       "php tests/unit/css-value-splitter.php",
       "php tests/unit/background-image-extractor.php",
+      "php tests/unit/cover-block-factory.php",
       "php tests/unit/cover-style-resolver.php",
       "php tests/unit/css-stylesheet-transformer.php",
       "php tests/unit/css-selector-matcher.php",
--- a/php-transformer/composer.json
+++ b/php-transformer/composer.json
@@ -61,6 +61,7 @@
       "php tests/unit/description-list-block.php",
       "php tests/unit/css-value-splitter.php",
       "php tests/unit/background-image-extractor.php",
+      "php tests/unit/cover-style-resolver.php",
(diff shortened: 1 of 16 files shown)


chubes4 (approved):
Thanks for filling this real PHP transformer coverage gap. The standalone core/cover save-shape implementation is well scoped, thoroughly tested, and does not introduce a runtime JavaScript dependency.

PR #764, opened 2026-08-02
Consume bounded runtime presentation evidence

### LINKED ISSUE #757: Consume runtime presentation evidence for post-load image wells
## Problem
Static artifact compilation sees authored HTML/CSS but not post-load runtime image transforms. In the Weebly Team surface, source JavaScript measures decoded image dimensions and applies a scale transform inside fixed clipped wells. The candidate has the same assets and bounds, but any static CSS substitute (`object-fit` or inferred transforms) uses a different Chromium compositor path and leaves 127,506 mismatch pixels.

## Direction
Define a generic optional browser-derived presentation evidence contract for artifact compilation. Evidence should bind source elements/assets to observed post-load computed image geometry and transform state with source hashes, viewport, browser revision, and lifecycle readiness. The compiler may consume proven evidence to project deterministic editable image presentation; static-only compilation remains unchanged.

## Acceptance
- Artifact schema can carry bounded runtime presentation observations for media wells.
- Observations include selector/identity, source asset hash, intrinsic and rendered dimensions, transform matrix/origin, clip bounds, viewport, browser revision, and readiness state.
- Compiler validates provenance before applying observations.
- Native/editable output consumes valid observations without fixture selectors.
- Missing/stale/mismatched evidence is diagnosed and ignored.
- Contract tests cover application and rejection.
- The mrfoxtalbot Team surface can be replayed through the runtime-evidence path.

## Evidence
Automattic/blocks-engine#746. r67: `/var/folders/lr/c_cmmt7s0592m4njz99v5yb40000gn/T/opencode/weebly-four-surface-direct-r67/visual-compare/mrfoxtalbot-weebly/team/`. Source and candidate asset hashes and visible bounds match; mismatch is isolated to source runtime compositor sampling.

## AI assistance
OpenAI gpt-5.6-sol through OpenCode analyzed r62-r68 evidence and drafted this issue. Chris Huber remains responsible for scope and implementation decisions.

chubes4:
## Summary
- define a generic, bounded artifact runtime presentation evidence schema
- bind observations to exact source HTML, body-relative image locators, and normalized asset hashes
- replay supported clip and axis-aligned scale/translation geometry on editable `core/image` output
- preserve deterministic static output for absent, stale, malformed, or unsupported evidence
- centralize artifact-root reference resolution and protect the internal marker namespace

## Verification
- `composer test`
- PHP lint for all modified PHP files
- `git diff --check`
- 258 PHP transformer parity fixtures
- package installation proof

Closes #757.

## AI assistance
OpenAI gpt-5.6-sol through OpenCode extracted the experimental implementation, reviewed and hardened the schema and replay path, added contract coverage, and ran verification. Chris Huber remains responsible for every line.

### DIFF
--- a/php-transformer/src/ArtifactCompiler/ArtifactCompiler.php
+++ b/php-transformer/src/ArtifactCompiler/ArtifactCompiler.php
@@ -138,7 +138,7 @@ final class ArtifactCompiler
             ),
         );
         if (array() !== $normalized['runtime_presentation_evidence']) {
-            $sourceReports['artifact']['runtime_presentation_evidence'] = array(
+            $sourceReports['artifact']['artifact_runtime_presentation_evidence'] = array(
                 'schema' => RuntimePresentationEvidence::SCHEMA,
                 'provenance' => $normalized['runtime_presentation_provenance'],
                 'observations' => $normalized['runtime_presentation_evidence'],
@@ -471,7 +471,8 @@ final class ArtifactCompiler
             );
         }
 
-        $result = ( new HtmlTransformer() )->transform($this->safeHtmlDocumentHtml($html, $sourcePath, $files), array(
+        $htmlForTransform = $this->safeHtmlDocumentHtml($this->withRuntimePresentationEvidenceMarkers($html, $sourcePath, $runtimePresentationEvidence), $sourcePath, $files);
+        $result = ( new HtmlTransformer() )->transform($htmlForTransform, array(
             'source'                    => $sourcePath,
             'source_scope'              => $sourceScope,
             'declarative_state_html'    => $html,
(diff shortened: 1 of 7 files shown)


chubes4:
Closing the branch, not the problem.

Issue #757 stays open and still wants this: `RuntimePresentationEvidence.php` is absent from trunk, so nothing superseded the work.

What is not worth keeping is this branch. It is 667 trunk commits behind, 238 of which touched `HtmlTransformer.php`, and it reports CONFLICTING/DIRTY. Its footprint in that file is +86/-4 — a small diff against a file that has been rewritten underneath it hundreds of times since 2026-08-02. Reconciling that costs more than writing it again against current trunk, and carries more risk of silently resurrecting stale behaviour.

Re-approach from #757 when it comes up.

## AI assistance

OpenAI GPT-5.6 Sol via OpenCode triaged the open PRs touching `HtmlTransformer`, confirmed #757 is still open and unsuperseded, and measured the branch drift. Chris Huber directed the triage.

Issue #762, opened 2026-08-01
Avoid paragraph flow around standalone block-participating inline leaves


chubes4:
## Problem

A standalone inline semantic element that participates as a block/layout child can be wrapped in a generated paragraph. Core paragraph margins then alter parent grid geometry even though the source leaf already owns its display and margin contract.

## Evidence

Fixture 89, transformer `d40ac8e2ef1504fcc69ca4a63d2283db11c17cba`, WP Codebox 0.18.4. Full passing artifacts: `/tmp/fixture-89-combined-v0184-round7-retry/artifacts`.

Source fallback cards contain a direct `<strong>` with `display:block; margin:12px 0 4.8px`. Candidate keeps `<strong>` but nests it in `p.wp-block-paragraph` with `margin:16px 0`.

- Source fallback grid height: 141.13px.
- Candidate: 156.33px.
- Exact local delta: +15.203px.

## Acceptance

- Preserve standalone block-participating inline leaves without generated paragraph flow margins.
- Retain source class/tag selector addressability, order, display, and authored margins.
- Keep ordinary phrasing content in native paragraphs.
- Preserve native/editor-valid output without core/html.
- Add rendered geometry coverage for direct strong/span leaves under grid/flex and normal prose negatives.

## AI assistance

OpenAI GPT-5.6 Sol using OpenCode analyzed complete source/candidate DOM and drafted this issue. Chris reviewed the evidence.

chubes4:
Round 10 isolates the remaining selector-addressability defect after the carrier geometry repair at combined transformer `f5e3ec74a9498e726dc9f049d48078505e8c17a1`.

Full-page source and candidate are both `1280x8067`; aligned mismatch is `31,263 / 10,325,760` pixels (`0.3028%`) with no page-level drift. The residual is entirely native inline descendants whose source DOM still matches the original selectors but whose projected author rules are absent:

- Four `.stage-output span` / `.stage-output strong` pairs lose block display, typography, line-height, and label margin. Each 260px card grows about 10px while remaining center-aligned.
- Four `.maintenance-loop li > span` markers lose their 30x30 grid circle, mono typography, colors, and radius, becoming stretched inline spans.

This extends the issue acceptance criterion for retaining source class/tag selector addressability: selector projection must recognize inline leaves serialized inside native Group/List output, including descendant and child combinators, when the candidate DOM remains selector-equivalent.

[...]

PR #737, opened 2026-08-01
Preserve author-owned layout children

chubes4:
Fixes #735

## Scope
Preserves author-owned external CSS layout behavior without allowing generated core Group defaults to take over layout ownership. The disclosed commit series preserves CSS-owned layouts, neutralizes semantic layout items, and projects `place-items` while keeping native block representations where valid.

## Verification
- `composer test` in `php-transformer` passed: canonical and unit contracts, 258 parity fixtures, and the packaging installation proof.
- Fixture 89 combined verification evidence: exact combined SHA `50309edbc004370de3f3c2468f013b0eee4fa1ce`; visual mismatch `0/10,325,760`; `170/170` native blocks; `340/340` editor observations; zero core/html fallbacks, warnings, and findings; dimensions `1280x8067`.
- The `f5e3ec74` fallback-card correction remains only in the local combined verification stack. Direct evidence shows it requires the separate inline-layout carrier series, so it is intentionally excluded from #735 ownership.

## AI Assistance
AI assistance: model gpt-5.6-sol, coding tool OpenCode, recreated the approved three-patch series with compliant disclosures and recorded verification evidence. Chris Huber remains responsible for every line.

Issue #735, opened 2026-07-31
Neutralize core flow defaults for author-owned layouts


chubes4:
## Problem

Source grid/flex containers retain classes but are serialized as Group blocks with `is-layout-flow` and `wp-block-group-is-layout-flow`. Core child margins/gaps then coexist with authored CSS, while generated paragraph/group wrappers alter direct-child topology. Fixture 37 amplifies this in `.ex-row`, nested Contact/Visit sections, header chrome, and other CSS-owned grids.

## Work

- Detect containers whose resolved author CSS owns grid/flex/spacing geometry.
- Preserve source semantic tag and direct-child topology where representable.
- Prevent core flow gap/margin defaults from contributing inside author-owned layouts.
- Emit diagnostics when Gutenberg save shape forces topology changes that invalidate structural selectors.

## Acceptance

- A five-cell authored grid retains child order/count and computed placement without injected flow spacing.
- Nested source sections retain semantic tags and authored padding.
- Ordinary Group flow layouts retain core defaults.
- Fixture 37 exhibition rows and Contact/Visit section geometry improve without fixture-specific CSS.

Evidence: Lab run `d17780c9-0100-4390-abe0-43eb11d465e7`; source `.ex-row`, `.form`, and global section rules. AI-assisted diagnosis: GPT-5.6 Sol via OpenCode; Chris owns acceptance.

chubes4:
Fixture `89-static-site-importer-architecture` adds a direct reproduction of this issue on commit `fe130c1b15c0cfa7e57dddde90b40ab4e7f7d152`.

Reproduction:

```bash
node tools/run-fixture-matrix.mjs \
  --static-site-importer /path/to/static-site-importer \
  --blocks-engine /path/to/blocks-engine \
  --target-fixture 89-static-site-importer-architecture \
  --batch-size 1 --concurrency 1 --local --no-visual-parity-gate
```

Homeboy run: `2d788144-6209-4b5c-b39e-d6798e2b49fe`

Observed result:

- 159 native blocks, zero `core/html`, and editor validity 100%.
- Source screenshot height: 8,067 px; WordPress candidate: 8,333 px.
- Aligned visual mismatch: 4,931,622 / 10,325,760 pixels (`47.76%`).
- Source `.hero-visual` computes one 474 px grid column and three 122 px card rows.
- Candidate `.hero-visual` is serialized as `wp-block-group is-layout-grid` and computes two 221 px columns with 214/294/267 px rows.
[...]

PR #773, opened 2026-08-02
Avoid paragraph flow around standalone block-participating inline leaves

chubes4:
## Root cause

Standalone inline semantic leaves that participate in block or layout flow were wrapped by generated paragraphs, letting core paragraph margins alter source-owned grid and flex geometry. The series keeps those leaves native and selector-addressable while neutralizing carrier flow.

Refs #762.

## Dependency and isolation

This PR targets `trunk` and contains no commits from #735. It may overlap the open #737 author-owned layout-child work in `HtmlTransformer`; review and merge order may require conflict resolution, but this series does not transplant #735.

## Verification

- Replayed the archived eight-commit series onto current `origin/trunk`; the rebuilt tree exactly matches the archived intended tree.
- `composer test` passed: contracts, unit suites, 258 PHP transformer parity fixtures, and package-install proof.
- Final combined evidence: `50309edbc004370de3f3c2468f013b0eee4fa1ce`; visual `0/10,325,760`; `170/170` native; `340/340` editor; zero core/html, warnings, and findings; `1280x8067`.

## AI assistance

gpt-5.6-sol via OpenCode audited the series disclosures and ancestry, replayed the verified patch series onto current trunk, ran tests, and prepared this PR. Chris Huber is responsible for every line.


chubes4:
Updated onto origin/trunk 1730461dcc99487a134e0346bb2ef94d1151e67b with merge commit 649790bfab5ec815ebd54f2c205b8accb7e3a846.\n\nConflict resolution retained trunk's native CSS-owned Group behavior and #778 list-item RichText-marker precedence, while preserving #762 standalone inline carriers, native control RichText marker survival, selector matching, and place-items projection. Added a focused coexistence assertion for standalone carriers and list-item markers.\n\nTests passed:\n- php tests/contract/run.php\n- php tests/unit/css-selector-matcher.php\n- php tests/unit/author-selector-semantics.php\n- php tests/unit/artifact-author-stylesheet-projection.php\n- composer test\n\nAI assistance: gpt-5.6-sol via OpenCode was used to diagnose, implement, test, and verify this change. Chris Huber remains responsible for every line.

PR #907, opened 2026-08-20
Fix heading RichText runtime serialization

### LINKED ISSUE #906: PHP transformer serializes heading RichText as a block attribute
## Problem

The PHP HTML-to-blocks transformer places a heading's source-derived RichText `content` in the working block's `attrs` array and also uses it to generate `innerHTML`. Passing that block to WordPress `serialize_blocks()` writes the source value into the block delimiter:

```html
<!-- wp:heading {\"content\":\"Event notifications\",\"level\":2} -->
<h2>Event notifications</h2>
<!-- /wp:heading -->
```

Normal Gutenberg serialization keeps heading content in saved HTML and omits it from the delimiter. The duplicate attribute changes runtime behavior on current WordPress/Gutenberg: when `WP_Block_Type::prepare_attributes_for_render()` receives explicit heading `content`, it passes the block schema's `type: rich-text` to `rest_validate_value_from_schema()`, producing `_doing_it_wrong` notices because `rich-text` is not a REST JSON Schema type.

Upstream context: WordPress/gutenberg#72180 and WordPress/wordpress-develop#11985.

## Production evidence

On a WordPress 7.0.4 multisite with Gutenberg 23.0.1:

- 12 transformed documentation posts contain explicit `core/heading.content` attributes.
- Rendering those posts through `the_content` produces 57 notices, all traced to `core/heading`.
- Rendering ten normally authored posts produces zero notices.

The affected content was generated by the transformer and persists the same duplicate heading shape shown above.

## Expected

Use heading `content` while constructing and analyzing the working block, but omit it from the block passed to WordPress serialization and rendering. Preserve real delimiter attributes such as heading `level`, presentation attributes, and block-support attributes.

## Acceptance criteria

- Heading HTML conversion serializes as `<!-- wp:heading {\"level\":2} -->...` without `content`.
- Heading text remains in saved HTML unchanged.
- Nested headings are normalized recursively.
- Native and fallback serialization follow the same contract.
- Native rendering does not receive the duplicate heading `content` attribute.
- Transformer working arrays are not mutated by runtime canonicalization.

chubes4:
## Summary
- canonicalize runtime copies of transformed block trees before serialization and rendering
- omit HTML-sourced `core/heading.content` while preserving heading text in saved HTML and keeping transformer working arrays unchanged
- cover nested fallback serialization plus native WordPress serialization and rendering paths

## Why
The PHP transformer currently duplicates heading RichText into the block delimiter and saved HTML. WordPress then validates the explicit `content` attribute using the block-only `rich-text` schema type, producing high-volume `_doing_it_wrong` notices. Normal Gutenberg markup derives heading content from saved HTML and does not persist it in delimiter attributes.

Closes #906.

## Testing
- `composer test`
- 278 parity fixtures
- canonical, unit, WordPress stub, and packaging contracts

Composer emits existing environment deprecation notices, but all gates pass.


php-transformer/src/WordPress/Runtime.php
```diff
@@ -5,6 +5,21 @@
...
+     * builds markup and reports. WordPress derives them from innerHTML when it
+     * parses a block, so passing them to serialize_blocks() duplicates the
+     * source value in the delimiter and exposes the block-only `rich-text`
+     * type to REST schema validation during rendering.
+     *
+     * @var array<string, array<int, string>>
+     */
+    private const HTML_SOURCED_RICH_TEXT_ATTRIBUTES = array(
+        'core/heading' => array( 'content' ),
+    );
```
mcsf, line 21: Why hardcode the case of heading content? Why not look at the block type and honour all sourced attributes (which by the way may not be of type rich-text)?

```sh
jq -r '.attributes[]?.source // empty' packages/block-library/src/*/block.json | sort | uniq -c
```
```
  37 attribute
   3 html
   6 query
   3 raw
  21 rich-text
```

chubes4: That would be a better approach for sure.

chubes4: Addressed this with schema-driven source attribute canonicalization in #911. It now uses live registered block schemas for all source kinds and a packaged core metadata snapshot in standalone mode. Thanks for catching the overly narrow map.

AI assistance: GPT-5.6-sol and GPT-5.6-terra via OpenCode and Homeboy were used to investigate, implement, and verify the follow-up. Chris Huber reviewed the changes and is responsible for every line.

PR #939, opened 2026-08-21
Preserve fallback diagnostics across pattern recursion

### LINKED ISSUE #919: Preserve fallback diagnostics across pattern recursion
## Problem

Registry-driven patterns recurse through `PatternContext` callbacks whose implementations create local fallback arrays and discard them. Accordion panel conversion and Navigation hoisted-element conversion can therefore produce blocks through `captureUnsupported=true` without returning the corresponding fallback evidence to the parent transform. Manually dispatched patterns such as Columns pass the real accumulator by reference, so the two recognition paths have different diagnostic semantics.

## Scope

- Introduce a structured pattern result carrying the matched block, fallbacks/findings, consumed source nodes, and any pattern-owned provenance.
- Make recursive child/element conversion return or merge its diagnostics explicitly.
- Preserve deterministic ordering and deduplication of findings.
- Apply the contract to current registry recognizers without broad pattern migration in this issue.

## Acceptance criteria

- A registry pattern containing a recursively preserved unsupported child returns the same fallback evidence as direct conversion.
- Probe-only matching remains side-effect free.
- Pattern matches cannot silently discard runtime-island or fallback findings.
- Existing parity and finding-contract tests remain green.

Related: #242.

## AI assistance

OpenAI GPT-5.6 Sol via OpenCode was used to audit pattern recursion and draft this issue. Chris Huber directed the investigation and remains responsible for the proposal.

chubes4:
## Summary
- add structured pattern results with explicit fallback propagation
- make rejected recognizer attempts transactional and side-effect free
- keep probe contexts mutation-free

## Verification
- focused contract regression passed
- full `composer test` passed before rebase, including 278 parity fixtures and packaging
- current trunk independently stops in `tests/contract/template-surfaces.php` on the shared Figma front-page assertion

Closes #919.

## AI assistance
OpenAI GPT-5.6 Sol via OpenCode general coding subagents implemented and reviewed this change. Chris Huber directed the work and remains responsible for every line.

### DIFF
--- a/php-transformer/src/HtmlToBlocks/Diagnostics/FallbackEmitter.php
+++ b/php-transformer/src/HtmlToBlocks/Diagnostics/FallbackEmitter.php
@@ -126,6 +126,28 @@ final class FallbackEmitter
     }
 
     /**
+     * @return array{runtime_island_origins: array<int, array{document: int, path: string}>, generated_block_names: array<string, string>, generated_block_identities: array<string, string>}
+     */
+    public function mutableState(): array
+    {
+        return array(
+            'runtime_island_origins'     => $this->runtimeIslandOrigins,
+            'generated_block_names'      => $this->generatedBlockNames,
+            'generated_block_identities' => $this->generatedBlockIdentities,
+        );
+    }
+
+    /**
+     * @param array{runtime_island_origins: array<int, array{document: int, path: string}>, generated_block_names: array<string, string>, generated_block_identities: array<string, string>} $state
+     */
+    public function restoreMutableState(array $state): void
+    {
+        $this->runtimeIslandOrigins     = $state['runtime_island_origins'];
+        $this->generatedBlockNames      = $state['generated_block_names'];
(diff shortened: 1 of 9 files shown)


chubes4:
Closing as superseded.

Issue #919 is closed as COMPLETED, and the closer was #1080 (merged 2026-08-23T03:04:27Z, one second before #919 closed). That PR solved fallback preservation across pattern recursion through `PatternRecursiveConverter`.

This branch took a different route via a new `PatternResult` type, which is absent from trunk — so its design did not land and the problem it targeted is already fixed.

The branch is also 457 trunk commits behind, 172 of which touched `HtmlTransformer.php`, and it currently reports CONFLICTING/DIRTY. There is nothing here worth reconciling against that.

## AI assistance

OpenAI GPT-5.6 Sol via OpenCode triaged the open PRs touching `HtmlTransformer`, traced #919 to its actual closer, and confirmed `PatternResult` never reached trunk. Chris Huber directed the triage.

PR #1265, opened 2026-08-27
fix(php-transformer): scan inline styles without PCRE backtracking

### LINKED ISSUE #1241: Preserve author stylesheets when inline CSS exhausts the PCRE backtracking limit
## Symptom

Importing https://www.360chiro.co.uk/ (Wix, 16 routes) through Studio `create --from` produced a site where 13 pages render with full layout parity but the homepage and `new-patients` render as bare document flow: no per-page `stylesheet-bundle-*.css` assets were emitted for those two routes, while every other route received its desktop+mobile bundle pair. The compile reported **no diagnostic** for either page.

## Root cause

`ArtifactNormalizer::withInlineStyleFiles()` extracts inline styles with:

```php
preg_match_all('@<style\b([^>]*)>(.*?)</style>@is', $content, $matches, PREG_SET_ORDER)
```

The lazy `(.*?)` across multi-megabyte style bodies exhausts PCRE's default `pcre.backtrack_limit` (1,000,000). `preg_match_all` then returns `false`, and the surrounding condition treats that **failure** exactly like "this file has no style tags" — the page is silently skipped, so no `inline-style` records exist, so no author stylesheet bundles are coalesced, so the page materializes unstyled.

Self-contained reproduction (PHP 8.x, default ini):

```php
$html = '<html><head><style>' . str_repeat('.a{color:#000}', 100000) . '</style></head><body></body></html>';
var_dump(preg_match_all('@<style\b([^>]*)>(.*?)</style>@is', $html, $m)); // bool(false)
var_dump(preg_last_error_msg()); // "Backtrack limit exhausted"
```

Real-world confirmation: on the captured artifact, `website/index.html` (2.65MB single style tag) and `website/new-patients/index.html` (1.46MB style tag) both fail with "Backtrack limit exhausted"; `website/fees/index.html` (max 207KB per tag) extracts 26 blocks cleanly. Wix inline CSS payloads over ~1MB per tag are common (`builder-components-css_*`).

## Affected call sites

The same backtracking-prone pattern appears in five places:

- `php-transformer/src/ArtifactCompiler/ArtifactNormalizer.php:400` (`withInlineStyleFiles`) — proven data loss above
- `php-transformer/src/ArtifactCompiler/ArtifactCompiler.php:2470` (`stylesheetAssetsForSource`) — a `false` here drops linked stylesheet assets too
- `php-transformer/src/HtmlToBlocks/HtmlTransformer.php:2749`
- `php-transformer/src/HtmlToBlocks/HtmlTransformer.php:2827`
- `php-transformer/src/HtmlToBlocks/Support/SvgMaterializationTrait.php:555`

## Suggested fix

Replace the lazy-regex extraction with a shared linear scanner (`stripos` for `<style` / `</style>`, no backtracking). A prototype extracts all 28/36 style blocks from the failing pages in ~1ms and is byte-identical to the regex output on pages where the regex succeeds. At minimum, a `preg_match_all` returning `false` must be distinguished from "no matches" and surfaced as a diagnostic rather than silent skip.

PR to follow.

ellenbauer:
## Problem

Compiling a captured page whose HTML carries a large inline `<style>` tag (>~1MB) silently loses every author stylesheet for that page: no `inline-style` records are materialized, no stylesheet bundles are emitted, and the page renders as bare document flow — with no diagnostic. On the #1241 reproduction (360chiro.co.uk, Wix, 16 routes), 13 pages styled correctly while the two pages carrying 2.65MB / 1.46MB style tags materialized unstyled. Wix captures routinely carry 1–3MB inline style payloads.

## Root cause

`preg_match_all('@<style\b([^>]*)>(.*?)</style>@is', ...)` exhausts PCRE's default backtracking budget (`pcre.backtrack_limit` = 1,000,000) once a single style body approaches ~1MB and returns `false`. Every call site treated `false` exactly like "no matches":

- `ArtifactNormalizer::withInlineStyleFiles()` — the page is skipped, so no inline-style assets exist (the proven production loss)
- `ArtifactCompiler::stylesheetAssetsForSource()` — ALL stylesheet assets for the page are dropped, linked CSS included, because the style/link alternation regex fails as a whole
- `HtmlTransformer::combinedAuthorStylesheet()` and `HtmlTransformer::inlineStylesheetPayloads()`
- `SvgMaterializationTrait::cssCustomProperties()`

## Fix

One shared linear scanner, `HtmlToBlocks\Style\StyleTagScanner` (a `stripos`/`strpos` walk, no backtracking), now backs all five sites:

- `styleBlocks()` — (attributes, css) pairs per style tag, verbatim capture
- `styleAndLinkTags()` — style and link tags in interleaved document order, preserving the cascade order `stylesheetAssetsForSource()` depends on (`<link\b[^>]*>` offsets merged against style spans; that pattern cannot backtrack pathologically)

Semantics are preserved for well-formed input: case-insensitive tags, attributes read up to the first `>`, verbatim bodies, an unclosed `<style>` yields no match, and a `<link>` inside a style body is not emitted. A 3,000-case randomized parity fuzz against the old regex found no divergence on well-formed markup; the one intentional divergence is that longer tag names (e.g. a `<style-guide>` custom element) are no longer treated as style openers, which the regex's `\b` wrongly accepted.

## Test evidence

`php-transformer/tests/unit/style-tag-scanner.php`, registered in the `test:unit` script:

- scanner/regex parity on well-formed markup, interleaved style+link document order, unclosed tags, case-insensitivity
- regression at integration level: a page with a ~1.2MB style tag followed by a small one first asserts the fixture really exhausts the default backtrack limit (`PREG_BACKTRACK_LIMIT_ERROR`), then compiles through `ArtifactCompiler::compile` and asserts both tags materialize (`index.inline-1.css`, `index.inline-2.css`) with their content present in emitted assets — under default pcre ini values

On trunk the integration section fails (zero stylesheet assets emitted for the page). With the fix, block output for the oversized page is byte-identical to the same page carrying a small style tag, and assets hold exactly one copy of the oversized CSS.

Local verification: `composer validate --strict`; the new test plus `responsive-document-variants`, `artifact-author-stylesheet-projection`, `compile-fragment-author-stylesheet`, and `artifact-normalizer-source-budget` all pass under `php -d memory_limit=1G`; parity (289 fixtures) and packaging pass. The only local `composer test` failure is the pre-existing 128M memory exhaustion in `html-transformer-shared-analysis-cache.php`, which fails identically on clean trunk and passes with `memory_limit=1G`.

## Out of scope / follow-up

A few other `<style` regexes remain (`ArtifactCompiler::withoutMaterializedStyleTags()` and the engine-support CSS collection, `ResponsiveDocumentVariants::styles()`, `TypographyParityAnalyzer`, `AssetReferenceCanonicalizer`). The block-materialization path was verified unaffected on oversized pages (byte-identical output as above); the rest degrade analysis or reference-rewriting quality on oversized bodies rather than dropping page assets, and can move to the scanner in a follow-up.

Fixes #1241

### DIFF
--- a/php-transformer/composer.json
+++ b/php-transformer/composer.json
@@ -86,6 +86,7 @@
       "php tests/unit/media-text-pattern.php",
       "php tests/unit/cover-style-resolver.php",
       "php tests/unit/css-stylesheet-transformer.php",
+      "php tests/unit/style-tag-scanner.php",
       "php tests/unit/admin-bar-accommodation.php",
       "php tests/unit/css-rule-analyzer.php",
       "php tests/unit/css-selector-matcher.php",
--- a/php-transformer/src/ArtifactCompiler/ArtifactCompiler.php
+++ b/php-transformer/src/ArtifactCompiler/ArtifactCompiler.php
@@ -16,6 +16,7 @@ use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\HtmlTransformerAnalysisC
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\AdminBarAccommodation;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\CssStylesheetTransformer;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\FormLayoutGraphBuilder;
+use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\StyleTagScanner;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\ShellLandmarkPolicy;
 use Automattic\BlocksEngine\PhpTransformer\Path\ArtifactPath;
 use Automattic\BlocksEngine\PhpTransformer\StaticSite\MaterializationPlanBuilder;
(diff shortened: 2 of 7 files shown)


ellenbauer:
Closing: superseded by #1255, which merged while this was in flight and fixed the same five call sites with the same linear-scanner approach (and closed #1241). No residual delta here worth keeping — the remaining backtracking-prone style regexes (withoutMaterializedStyleTags, the engine-support CSS collection, ResponsiveDocumentVariants::styles(), TypographyParityAnalyzer, AssetReferenceCanonicalizer) were out of scope for both PRs and could be a small follow-up if desired.

Issue #1279, opened 2026-08-27
Wrapper-shaped children inflate repeatable_children in SubtreeClassifier


chubes4:
## Problem

`SubtreeClassifier::repeatableChildCount()` signatures a subtree's immediate children by tag name and class list only:

```php
$classes = $this->classNames($child);
sort($classes);
$signature = strtolower($child->tagName) . '|' . implode('.', $classes);
```

The child's contents are never considered. Sources that wrap every content element in an identical wrapper therefore report structural repetition that the content does not have. A section heading and the body copy beside it collapse to one signature.

`repeatable_children === 2` adds `+2.0` toward `BUCKET_CUSTOM_BLOCK`, and `FallbackEmitter::generatedBlockStem()` names a `collection` at `>= 2`, so a plain two-part section can score toward the generated fallback path and be labelled a collection.

## Reproduction

```php
$section = $element(
    '<div role="group" class="section">'
    . '<interact-element><div class="rich"><h1><span>Commute with Purpose</span></h1></div></interact-element>'
    . '<interact-element><div class="rich"><p>Turn your daily transit into rewards.</p></div></interact-element>'
    . '</div>'
);
$classifier->classify($section, new ClassificationContext('.section { display: grid; }', ''))
    ->signals()['repeatable_children'];
// => 2
```

Both wrappers are class-less, so both produce the signature `interact-element|`. A heading and a paragraph are counted as a repeating pair.

The same holds for headings at different levels in matching wrappers:

```php
'<div class="stack">'
. '<div class="row"><h2>Section title</h2></div>'
. '<div class="row"><h4>Sub label</h4></div>'
. '</div>'
// repeatable_children => 2
```

## Scope

This is a false positive in the repetition signal itself. It is independent of #1278: for the Wix source in that issue the affected subtrees reach `FallbackEmitter::maybeGenerateCustomBlock()` with `$confirmedComponent = true`, which bypasses `SubtreeClassifier` entirely. A change to `repeatableChildCount()` produced a byte-identical import for that source. The two defects are unrelated and should be fixed independently.

## Proposed fix

Pair the child's tag and class shape with the heading levels it carries. Peers in a collection agree on heading structure; a heading and adjacent body copy do not, even when their wrappers match exactly.

[...]

PR #1281, opened 2026-08-27
Unwrap wrapper-convention custom elements before component promotion

### LINKED ISSUE #1278: Safe custom elements are promoted to companion blocks regardless of native representability
## Problem

`SubtreeClassifier::repeatableChildCount()` signatures a subtree's immediate children by tag name and class list only. It never inspects the child's contents. Sources that wrap every content element in a consistent wrapper element therefore report structural repetition that does not exist, and ordinary page sections fall off the component decision ladder into generated fallback companion blocks.

A heading and a paragraph sitting in two identical wrappers are scored the same as a genuine repeating collection.

## Reproduction

Source: `https://www.wix.com/demone2/nimbus-commute` (single-page Wix demo site)

Pipeline: Data Liberation capture -> Blocks Engine `trunk` -> Static Site Importer `main`.

The homepage hero is semantically a heading, a subheading, and a button. It materialized as a single `custom/collection-*` generated block carrying 104,548 bytes of source markup in a `content` string attribute.

The captured hero subtree has this shape:

```html
<div data-testid="internal-container-content" role="group" class="comp-... max-width-container">
  <interact-element data-interact-key="...">
    <div class="... wrichtext ..."><h1><span>Commute<br>with Purpose</span></h1></div>
  </interact-element>
  <interact-element data-interact-key="...">
    <div class="... wrichtext ..."><p>Turn your daily transit into tangible rewards for the planet and your wallet</p></div>
  </interact-element>
  ...
</div>
```

Measured immediate-child signatures for that subtree:

```
interact-element|                                 x2
div|_WRichTextNewSkin_vcky1_163 _support...       x1

repeatable_children = 2
```

`<interact-element>` carries no `class`, so every wrapper collapses to the identical signature `interact-element|`.

## Mechanism

1. `repeatableChildCount()` builds `strtolower($child->tagName) . '|' . implode('.', $classes)` and returns the maximum count. The child subtree is not considered.
2. Classless wrapper elements collapse to one signature, so `repeatable_children = 2`.
3. `SubtreeClassifier` scoring adds `+2.0` to `BUCKET_CUSTOM_BLOCK`:

```php
} elseif ( $s['repeatable_children'] === 2 ) {
    $scores[self::BUCKET_CUSTOM_BLOCK] += 2.0;
}
```

4. `FallbackEmitter::generatedBlockStem()` returns `collection` at `repeatable_children >= 2`, which matches the observed `custom/collection-*` block names.
5. Once confidence reaches `GENERATION_CONFIDENCE_THRESHOLD` (0.7), `FallbackEmitter::maybeGenerateCustomBlock()` freezes the whole subtree into a `content` string attribute.

## Impact on this source

Across the materialized homepage:

- 22 of 24 generated companion blocks are `collection-*` or `gallery-*` fallbacks.
- 578 KB of 700 KB of serialized block markup (82.6 percent) is opaque HTML inside `content` attributes.
- The rendered page shows 232 visible words while native blocks account for only 94, so about 60 percent of page text is not represented as blocks.
- The page reports zero invalid blocks and zero `core/html` fallback, so existing quality signals report success.

The generated block's editor UI is a `RawHTML` preview plus a textarea, so these sections are not meaningfully editable.

This is the failure mode described in #868, reached by a simple landing page with no exotic components.

## Inconsistent thresholds

`$isContentUnit` requires `repeatable_children >= 3`:

```php
$isContentUnit = $s['repeatable_children'] >= 3
    || $s['media_gallery']
    || $s['interactive_role']
    || $s['cohesive_content_unit'];
```

but the `custom_block` score bumps at `=== 2` and `generatedBlockStem()` names a `collection` at `>= 2`. Two children are enough to be scored toward fallback and named a collection while not qualifying as a content unit.

## Why this is not source-specific

Any source that wraps content in a consistent wrapper element reproduces this: Wix `interact-element`, and comparable wrapper conventions from other site builders and component frameworks. The signal currently measures markup convention rather than semantic repetition.

## Proposed fix

Make the repetition signal reflect content shape rather than wrapper shape. Both directions are generic:

1. Descend through render-neutral wrappers before computing signatures. A classless custom element or `div` with a single element child is a wrapper, not a content unit. Related prior work exists in #1039 for media wrapper chains.
2. Include a coarse subtree fingerprint in the signature, for example the child's descendant tag profile, so `wrapper > h1` and `wrapper > p` do not collide.

## Acceptance criteria

1. A subtree whose children are identical classless wrappers containing structurally different content (heading vs paragraph) does not report inflated `repeatable_children`.
2. A genuine repeating collection (sibling cards with similar internal structure) still reports repetition and still classifies as a content unit.
3. The Nimbus hero maps to native blocks (`core/heading`, `core/paragraph`, `core/buttons`) rather than a generated `collection-*` fallback.
4. Coverage proves both the false-positive and true-positive cases.

## AI assistance

Investigated with Claude Sonnet 4.5 via OpenCode. The model ran the import, inspected the materialized database content, generated companion plugin renderers, and editor behavior, then traced the classification path through `SubtreeClassifier` and `FallbackEmitter` and measured the child signatures of the captured hero subtree. Chris Huber reviewed and is responsible for this report.

chubes4:
Closes #1278.

## Problem

`collectGeneratedComponentCandidates()` let any tag containing a hyphen bypass the repeated-child-structure requirement:

```php
&& ($this->hasRepeatedDirectChildTags($element) || str_contains(strtolower($element->tagName), '-'))
```

A custom element tag was therefore treated as sufficient evidence that a subtree needed a generated block. Promotion then ran with `$confirmedComponent = true`, which skips `SubtreeClassifier` entirely, so no scoring could intervene.

Sources that wrap ordinary content in presentation-only custom elements had headings, copy, and quotes frozen into opaque generated blocks. For the Wix source in #1278, 24 of 28 generated blocks were rooted at `<interact-element>`, including one wrapping nothing but the page `<h1>`.

## Approach

Remove the wrappers before promotion rather than teaching promotion to recognise them. Once the wrapper is gone, the content converts through the existing native path with no special-casing.

A custom element tag is read as a wrapping convention when it recurs across the document while the content it wraps diverges. A real component repeats its internal structure; a wrapper does not. Structure alone cannot separate the two, because both commonly wrap a single child.

Instances are unwrapped only when they carry no identity, behaviour, or presentation: no `id`, `class`, `role`, or `style`, no interactive attributes, not a runtime DOM target, and no motion structure token. That vocabulary matches the gates already used by `coalescedSingleGroupWrapper()`.

Author CSS is judged by render neutrality rather than presence. Sources declare exactly one rule for these wrappers:

```css
interact-element{display:contents}
```

`display:contents` states the element generates no box and renders as its children, so a presence check preserved the wrapper on the strength of the rule that proves it is safe to remove. The pass also reads the document's embedded styles alongside the declared stylesheet so it observes the same CSS as the rest of the transform.

The pass runs before author-selector preparation so style projection and component promotion both observe content rather than the wrapping convention.

## Result on the reproduction source

Measured end to end by rebuilding the Static Site Importer development package against this branch and re-running the import.

| Measure | Before | After |
| --- | --- | --- |
| Opaque HTML in generated `content` attributes | 578,428 bytes (82.6%) | 2,586 bytes (0.8%) |
| Generated companion blocks on the page | 16 | 2 |
| Serialized block markup | 700 KB | 308 KB |
| Words represented in native blocks | 94 | 450 |
| `core/heading` | 0 | 26 |
| `core/image` | 6 | 38 |
| `core/quote` | 0 | 4 |

Transformer-level count for the same input drops from 17 generated blocks to 4.

Inline SVG artwork that previously disappeared now renders, because the artwork reaches the frontend through native blocks instead of the companion renderer's `wp_kses_post()` boundary (see static-site-importer#1361).

## Verification

- `composer test` passes end to end (exit 0).
- `tests/unit/custom-block-generator.php`: 58 passed, including `6: a deep safe custom-element host becomes one exact companion block`. `<fluid-card-grid>` is used once and keeps a repeated internal shape, so it is not read as a convention and still becomes a companion block.
- `tests/unit/inert-capture-scaffolding.php`: 11 assertions passed, including the `fluid-columns-repeater` case.

## Effects verified against the pre-change build

Each item below was measured by importing the same source with and without this change and comparing the rendered pages, including aligned screenshots and DOM geometry.

### Restored by this change

The signup form now exists. Before this change the page contained zero `input` elements: the form lived inside a generated companion block, and that renderer's `wp_kses_post()` boundary strips form controls exactly as it strips SVG (static-site-importer#1361). After this change the page has six `input` elements, three of them laid out and visible.

Inline SVG artwork also returns for the same reason: numbered feature-card icons render where they were previously empty boxes, and section headings render as text.

### Not a regression: footer logo and legal links

An earlier note on this PR claimed the footer lost its logo and `Privacy Policy` / `Accessibility Statement` links. That was a measurement error. The after page is 505px taller (5804 → 6309), and a bottom-of-page crop compared the original footer to empty space below the after footer.

Aligned to the Privacy Policy y-coordinate, both builds match the original's x-position (281 / 446). The footer wordmark is present in both DOMs at 112×28. The legal links are present in both. Nav and social labels wrap mid-word in both builds; that wrapping is pre-existing.

The after footer group is taller (431px vs 123px), which is leftover empty space from more of the page now participating in flow, not missing content.

### Pre-existing, unchanged by this change

Verified identical before and after:

- The header navigation pill is misplaced and overlaps the first section (`nav` at x=571, y=33, 298×29 and `header` at 1440×235 in both builds).
- A secondary navigation label renders one letter per line.
- Footer social labels wrap mid-word.

### Remaining defects on the restored form

- The three fields render side by side (x = 67, 227, 387 at a shared y) where the source stacks them vertically.
- Field labels are carried only as `aria-label` and are not rendered visibly.
- The submit control renders as a full-width bar labelled `Button` rather than the source's pill labelled `Claim My Spot`.

This source is not solved. The remaining items are separate layout and pattern-recognition concerns, not blockers for the promotion-path change in this PR.

## AI assistance

Claude Sonnet 4.5 via OpenCode traced the promotion path, produced the measurements, implemented the pass, and ran verification including repeated end-to-end imports to confirm effect. Two earlier root-cause hypotheses were disproved by measurement before this one; that history is recorded on #1278 and #1279. Chris Huber reviewed and is responsible for the submitted change.

### DIFF
--- a/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
+++ b/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
@@ -88,6 +88,17 @@ use DOMNode;
 final class HtmlTransformer
 {
     private const GENERATED_COMPONENT_MIN_SOURCE_DEPTH = 14;
+
+    /**
+     * Instances a custom element tag must reach before recurring usage can be
+     * read as a wrapping convention rather than a component.
+     */
+    private const WRAPPER_CONVENTION_MIN_INSTANCES = 3;
+
+    /**
+     * Depth of the shallow tag skeleton compared across wrapper instances.
+     */
+    private const WRAPPER_CONTENT_SHAPE_DEPTH = 2;
     use ButtonLinkDispatchTrait;
     use DomHelpersTrait;
     use ElementConversionTrait;
@@ -522,6 +533,10 @@ final class HtmlTransformer
 
         $this->navigationBlockNormalizer->hydrateDuplicateSubmenus($body);
         $this->materializeDeclarativeCounters($body, (string) ($options['declarative_state_html'] ?? ''));
+        // Remove wrapper-convention custom elements before author selectors are
+        // prepared, so style projection and component promotion both observe the
+        // content rather than the source's wrapping convention.
+        $this->unwrapRenderNeutralCustomElements($body, (string) ($options['static_css'] ?? ''));
         $this->prepareAuthorSelectorSemantics($html, (string) ($options['static_css'] ?? ''), $body, $options);
         $this->fallbackEmitter()->configure($this->transformationProvenance()->fallback(), $this->runtimeBehavior()->runtimeScriptMetadata(), $this->runtimeSelectors(), $this->authorSelectorProjections()->tagMarkers());
         // Author-selector preparation marks source nodes for later projection.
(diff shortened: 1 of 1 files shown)


chubes4:
Correction on the footer: it is not a content-loss regression.

The after page is 505px taller (5804 → 6309). A bottom-of-page crop compared the original footer to empty space below the after footer, which made the logo and legal links look missing.

Aligned to the Privacy Policy y-coordinate:

- `Privacy Policy` / `Accessibility Statement` are present in both builds at the original x (281 / 446)
- The footer wordmark is present in both DOMs at 112×28
- Nav/social wrapping is identical before and after

The after footer group is taller (431px vs 123px). That is leftover empty space from more of the page now participating in flow, not dropped content.

Sorry for the false alarm. The PR body is updated.

## AI assistance

Claude Sonnet 4.5 via OpenCode compared DOM geometry and aligned screenshots across original, pre-change, and post-change imports. Chris Huber reviewed.

Issue #1197, opened 2026-08-25
Navigation author-rule collection exhausts memory on large stylesheets


chubes4:
## Problem

Compiling a prepared page with a large shared author stylesheet exhausts a 512 MB PHP limit before the first page receipt is written. The failure occurs in navigation compatibility CSS collection while walking selector specificity.

## Reproduction evidence

- Input corpus: 86 HTML documents with 89.2 MB inline CSS, 93.1% duplicate CSS bytes.
- The bounded importer successfully prepares all 86 page plans under 512 MB.
- The first five-page compile batch terminates before writing any receipt.
- Fatal path: `HtmlTransformer::navigationAuthorStyleRules()` -> `collectNavigationAuthorStyleRules()` -> `navigationSelectorSpecificity()`.
- At failure, collection had reached source order 2457.
- Error: `Allowed memory size of 536870912 bytes exhausted`.

The collector copies the complete stylesheet into `$directCss`, repeatedly replaces full at-rule spans, retains nested CSS/event copies, then materializes all regex captures before visiting rules. Peak memory therefore scales as multiple complete stylesheet copies rather than the active rule.

## Expected

Navigation author-rule collection should visit style rules incrementally, preserve source order and conditional ancestry, and keep peak memory bounded by the source stylesheet plus the active rule and retained relevant facts.

## Acceptance

- Existing navigation cascade and conditional-state contracts remain green.
- A large stylesheet regression fixture proves collection completes under a deterministic memory bound.
- The 512 MB corpus compile advances past its first receipt.

## AI assistance

OpenAI GPT-5.6 Sol via OpenCode traced the runtime stack, inspected the collector, and drafted this report. Chris Huber directed the investigation and remains responsible for the report.

Issue #242, opened 2026-06-27
[Epic] Refactor for maintainability + parallel-safe swarm: decompose HtmlTransformer, formalize finding contract, modularize matrix


chubes4:
## Why now
`HtmlTransformer.php` is a ~3k-line god-object (pattern dispatch + menu-toggle suppression + button/style resolution + fallback diagnostics + semantic parity + runtime islands). Every transformer change this session had to edit it, forcing agents to be *serialized* instead of parallel. Decomposing it is the **precondition for the mass pattern-recognition swarm** (modular files → parallel agents → no collisions) and just good maintenance.

## Safety net
128 parity fixtures + the contract suite make **behavior-preserving** refactors verifiable: identical green before/after, zero block-output diff. Every refactor slice must prove this.

## Workstreams
1. **Decompose `HtmlTransformer`** into focused modules — `Patterns/` (exists) + a `Diagnostics/` layer (fallback / behavior-loss / semantic-parity emission) + a `Style/` layer (generalize `ButtonStyleResolver` into element-agnostic CSS→native-attribute resolution) + nav-chrome / runtime-island units. `HtmlTransformer` becomes a thin orchestrator. (multi-slice)
2. **Formalize the cross-repo finding/diagnostic contract** — the finding schema is implicit across blocks-engine (FallbackDiagnostic / RuntimeDependencyParityReport / VisualParity), SSI (loss-classes / report-diagnostics), homeboy-rigs (matrix acceptance). Make it one explicit versioned contract.
3. **Modularize `fixture-matrix.mjs`** (homeboy-rigs) — split accreted freshness / acceptance / editor-validation / visual-parity / metrics steps into composable step + collector modules.

## Rules
- Behavior-preserving; green parity+contract before and after each slice.
- Refactor on a QUIET base — never race feature/swarm agents on the same files.
- Pure extraction/move per slice; no logic changes smuggled in.

chubes4:
Current architecture audit update, based on `trunk` at `57e7b3f9` and the open registry completion in #1049:

- `HtmlTransformer.php` is now 15,512 LOC, not the approximately 3k LOC recorded when this epic was opened.
- Per-transform state moved into `HtmlTransformerSession`, but the transformer still accesses more than 120 mutable fields through `__get` / `__set` / `__isset`, so phase inputs and outputs remain implicit.
[...]

PR #1333, opened 2026-08-29
Convert NavigationStyleProjectionTrait into a collaborator (#1327)

### LINKED ISSUE #1327: Convert NavigationStyleProjectionTrait into a collaborator, and order the remaining workstream 1 slices by coupling (#242)
Slice 7 for #242, workstream 1. Follows #1310, #1312, #1315, #1316, #1317, #1318.

#1318 landed the `StyleResolver` collaborator and made an observation worth acting on: for `StyleResolutionTrait` (2,822 lines) vs `FormDispatchTrait` (1,865 lines), **size and coupling were inversely related** — the bigger trait had half the external dependencies.

That observation was made against two traits. This issue extends the same measurement to all five remaining single-consumer traits, so the rest of workstream 1 is ordered by evidence rather than by line count.

## Measurement

External dependency = a `$this->` symbol referenced inside the trait that the trait does not itself define. This is the surface that must become an explicit context when the trait becomes a collaborator, so it is the real cost driver. The method reproduces #1318's reported `FormDispatchTrait = 30` exactly.

| Trait | Lines | External deps | Lines / dep |
|---|---:|---:|---:|
| *`StyleResolutionTrait` (landed, #1318)* | *2,822* | *15* | *188* |
| **`NavigationStyleProjectionTrait`** | **1,319** | **12** | **110** |
| `NavigationToggleSuppressionTrait` | 781 | 10 | 78 |
| `FormDispatchTrait` | 1,865 | 30 | 62 |
| `SvgMaterializationTrait` | 983 | 20 | 49 |
| `ElementConversionTrait` | 594 | 89 | 7 |

`DomHelpersTrait` (604 lines, 4 consumers) is excluded — it is a legitimate shared mixin, not a single-consumer slab.

Current object scope: **12,711 lines in `HtmlTransformer` + 5,542 in the five single-consumer traits ≈ 18,250 lines**, down from roughly 22,000 before the recent slices.

## Proposed next slice: `NavigationStyleProjectionTrait`

Best remaining leverage, and the boundary is unusually clean.

**Twelve external deps:** `attr`, `authorStyles`, `elementSelector`, `generatedSupportStyles`, `materializedAssets`, `navigationSourceOwnershipClasses`, `parsedCssSelector`, `runtimeBehavior`, `safeAnchor`, `sourceStyles`, `styleResolver`, `transformationEvidence`.

**Ten entry points, every one reached only from `HtmlTransformer`:** `directNavigationSupportCss`, `materializeStylesheetAsset`, `materializeEditorStaticStateStylesheet`, `listNavigationInlineMarginRules`, `listNavigationPaddingRules`, `listNavigationItemAnchorRules`, `navigationItemStateAnchorRules`, `navigationColorInteractionStates`, `navigationLinkTextColorRules`, `sourceMobileNavigationOverlayBackground`.

No other trait or class reaches into it, so this is a single-consumer migration rather than #1318's 292-call-site fan-out across seven files.

Suggested shape follows the established precedent — a `NavigationStyleProjector` behind a `NavigationStyleProjectionContext`, with per-transform state resolved through closures because the collaborator is built once per transformer but must observe the transform currently running.

## Ordering note for the remaining three

`NavigationToggleSuppressionTrait` (781 lines, 10 deps) is the natural follow-on. It shares only two external deps with `NavigationStyleProjectionTrait` (`attr`, `styleResolver`) and there are **zero cross-calls between the two navigation traits** in either direction — they are independent despite the shared subject matter. They can land as separate slices without one blocking the other.

## `ElementConversionTrait` is not a collaborator candidate

Calling this out so a future slice does not attempt it by size.

594 lines, but **89 external deps against only 5 owned methods** — `convertFigureElement`, `convertListElement`, `convertSvgElement`, `convertInlineContentElement`, `convertFlowContainerElement`. It is not a cohesive unit that was mixed in; it is five slabs of the `convertElement` dispatch chain lifted across a file boundary. Its ratio of 7 is an order of magnitude worse than anything else on the list.

Converting it to a collaborator would mean threading an 89-operation context, which is a service locator wearing a different hat. These five belong in the `Elements/` converter model instead — the one proven by `TextLeafElementConverter` (#1310), `TableElementConverter` and `UnsupportedElementRecorder` (#1315) — extracted one element type at a time with explicit inputs and returned outputs.

## Verification bar

Same as prior slices, per the epic's behavior-preserving rule:

- `serializedBlocks` SHA-256 plus block, diagnostic, and fallback counts captured across all **383** fixture HTML files on clean `trunk` before any edit, recaptured after. Required result: 0 differing, 0 throwing.
- `composer test` exit 0.
- Pure extraction — no logic changes smuggled into the slice.
- Trait deleted outright, no back-compat shim.

Fallback and diagnostic counts matter specifically here: this code materializes stylesheet assets and records transformation evidence, so identical counts are what demonstrate the diagnostic path is untouched.

---

*AI assistance disclosure: this issue was researched and drafted by Claude Sonnet 4.6 running in Claude Code, operated by @chubes4. The AI read `blocks-engine` at `trunk` `24b115ac`, computed the trait line counts, external-dependency counts, entry-point consumers, and cross-call analysis quoted above, validated its dependency-counting method against the figure independently reported in #1318, and wrote the writeup. No code was changed. Reviewed by a human before opening.*

chubes4:
Slice 7 for #242, workstream 1. Closes #1327. Follows #1310, #1312, #1315, #1316, #1317, #1318.

`NavigationStyleProjectionTrait` was a 1,319-line single-consumer mixin. Per #1318's finding that trait extraction moves code across a file boundary without changing the `$this` surface, it is now `NavigationStyleProjector` behind a 10-operation `NavigationStyleProjectionContext`, with no `$this` access to the transformer.

#1327 picked this trait by measuring lines-removed per external dependency rather than by size. It ranked highest of the five remaining (110 lines/dep vs `FormDispatchTrait`'s 62), and the boundary held up: 9 entry points, every one reached only from `HtmlTransformer`, so this was a single-consumer migration rather than #1318's 292-call-site fan-out.

## Two members moved in opposite directions

Neither was in the original plan; both came out of reading the actual call sites.

**`navigationSourceOwnershipClasses` moved into the projector.** It lived on `HtmlTransformer` but was reached only from this trait, and is pure string work with no `$this` dependency. Threading it through the context would have widened the surface to describe a method nothing else uses. It leaves the transformer entirely.

**`materializeStylesheetAsset` moved the other way, into `HtmlTransformer`.** It lived in the navigation trait, but three of its four call sites are engine-support and author stylesheets — it is transformer-owned asset materialization that happened to be filed under navigation. Moving it into a class named `NavigationStyleProjector` would have meant the transformer calling `$this->navigationStyleProjector->materializeStylesheetAsset()` to write its author CSS. It stays with the transformer; the projector reaches it through the context.

That removal is also why the context is 10 operations rather than the 12 dependencies #1327 measured: `materializedAssets` was used only inside `materializeStylesheetAsset`, so it left the surface with it.

## A constant the dependency analysis missed

`self::EMPTY_RUNTIME_TARGET_CLASS` is defined on `HtmlTransformer` and referenced from the trait. The `$this->`-based dependency counting in #1327 does not see `self::` constants, so this was invisible until the contract suite hit it.

It is now `public`, referenced as `HtmlTransformer::EMPTY_RUNTIME_TARGET_CLASS`, following the existing `ButtonLinkDispatcher::POSITIONED_FRAGMENT_LINK_CARRIER_CLASS` convention where the class that emits a marker owns the constant and readers reference it. Here the transformer emits the marker and the projector targets it in editor static-state CSS.

**Worth carrying into the remaining slices:** extend the dependency measurement to `self::`/`static::` constants before estimating a trait, not just `$this->`.

## Behavior preservation

`serializedBlocks` SHA-256 plus block, diagnostic, fallback, asset-count, asset-key and coverage fingerprints, captured across all fixture documents on the clean base and recaptured after:

```
383 documents — DIFFERING: 0 — threw: 0

                baseline    after
blocks             1,631    1,631
diagnostics        6,772    6,772
fallbacks          1,037    1,037
assets             3,239    3,239
```

`composer test` exit 0.

Two details about how that baseline was built, because they changed the result:

**Stylesheets are attached.** Navigation style projection reads author CSS, so a corpus run with no `static_css` would not exercise the code under change. The harness concatenates each fixture's stylesheets and passes them through `static_css`; 82 of 87 fixtures carry CSS.

**The harness was verified to reach the code before it was trusted.** All 9 entry points were temporarily instrumented and confirmed to fire within the first 60 documents, rather than assuming corpus breadth implies coverage.

## Where the corpus was not enough

The undefined-constant fault above did **not** surface in the 383-document run — it reported 0 throwing while the fault was live, because the branch holding that constant sits behind a conditional the corpus never enters. The contract suite caught it.

Recording this because these slices lean on the corpus as the primary behavior gate: it proves output identity across a wide input surface, but it is not a branch-coverage instrument. Corpus green plus suite green is the real bar; corpus green alone is not.

## Impact

| | trunk | after |
|---|---:|---:|
| `HtmlTransformer.php` | 12,711 | 12,771 |
| Named methods on the class | 515 | 516 |
| **Methods in the transformer's object scope** | **548** | **516** |
| **Object scope (class + single-consumer mixins)** | **18,253** | **16,994** |

The class grows 60 lines — the property, the constructor wiring, the context factory and the relocated `materializeStylesheetAsset` — while 1,259 lines and 32 methods leave the shared mutable scope. Same shape as #1318, which grew the file by 37 lines while removing 2,822 from the object scope.

Remaining single-consumer mixins, by the #1327 ranking: `FormDispatchTrait` (1,865 / 30 deps), `SvgMaterializationTrait` (983 / 20), `NavigationToggleSuppressionTrait` (781 / 10), `ElementConversionTrait` (594 / 89 — not a collaborator candidate, see #1327).

## Test updates

One test reflected on `navigationAuthorStyleRules()` via `HtmlTransformer` — the #1197 memory-bound contract. It now reaches the method through the transformer's collaborator, so it still exercises the real wiring including the context closure that resolves the running transform's session state. The method stays private on the projector and is reached by reflection rather than widening its API for a test, following #1318. A sweep for other tests referencing any of the 34 moved methods found none.

---

*AI assistance disclosure: implemented and drafted by Claude Sonnet 4.6 running in Claude Code, operated by @chubes4. The AI measured the trait's dependency surface, captured and verified the pre-change corpus baseline, performed the extraction, migrated the call sites, diagnosed the undefined-constant and moved-method-reflection failures surfaced by the suite, and wrote this description. Reviewed by a human before opening.*

### DIFF
--- a/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
+++ b/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
@@ -73,7 +73,8 @@ use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Patterns\SocialLinksPatt
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\AuthorSelectorProjectionState;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\AuthorStyleAnalysis;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\LayoutGeometryState;
-use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\NavigationStyleProjectionTrait;
+use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\NavigationStyleProjectionContext;
+use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\NavigationStyleProjector;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\PresentationResolutionCache;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\CssSelectorMatcher;
 use Automattic\BlocksEngine\PhpTransformer\HtmlToBlocks\Style\CssSelectorMatchCache;
@@ -113,7 +114,6 @@ final class HtmlTransformer
     use DomHelpersTrait;
     use ElementConversionTrait;
     use FormDispatchTrait;
-    use NavigationStyleProjectionTrait;
     use NavigationToggleSuppressionTrait;
     use SvgMaterializationTrait;
 
@@ -249,6 +249,8 @@ final class HtmlTransformer
 
     private readonly StyleResolver $styleResolver;
 
+    private readonly NavigationStyleProjector $navigationStyleProjector;
+
     private readonly RuntimeIslandAnalyzer $runtimeIslands;
 
     private readonly ButtonLinkDispatcher $buttonLinkDispatcher;
@@ -290,7 +292,12 @@ final class HtmlTransformer
 
     private const EMPTY_FLEX_ITEM_CLASS = 'blocks-engine-empty-flex-item';
 
-    private const EMPTY_RUNTIME_TARGET_CLASS = 'blocks-engine-empty-runtime-target';
+    /**
+     * Marks an emptied block that exists only as a runtime target. Emitted
+     * here and read back by {@see NavigationStyleProjector}, which projects the
+     * editor static-state CSS that hides it.
+     */
+    public const EMPTY_RUNTIME_TARGET_CLASS = 'blocks-engine-empty-runtime-target';
 
     private const CSS_OWNED_LAYOUT_CLASS = 'blocks-engine-css-owned-layout';
 
@@ -369,6 +376,10 @@ final class HtmlTransformer
         $this->textLeafConverter = new TextLeafElementConverter($this->createTextLeafElementContext());
         $this->richTextConverter = new RichTextElementConverter($this->createRichTextElementContext());
         $this->styleResolver = new StyleResolver($this->createStyleResolutionContext(), $this->analysisCache);
+        $this->navigationStyleProjector = new NavigationStyleProjector(
+            $this->createNavigationStyleProjectionContext(),
+            $this->styleResolver
+        );
         $this->runtimeIslands = new RuntimeIslandAnalyzer($this->createRuntimeIslandContext());
         $this->buttonLinkDispatcher = new ButtonLinkDispatcher($this->createButtonLinkDispatchContext());
         $this->tableConverter = new TableElementConverter($this->createTableElementContext());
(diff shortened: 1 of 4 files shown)


chubes4:
## Correction: the markup fingerprint in the description above was vacuous, and the re-run

While measuring the next slice I found a defect in the harness that produced this PR's evidence, and it invalidates one of the claims made above.

`TransformerResult::toArray()` emits **snake_case** keys. The harness read `$result['serializedBlocks']`; the actual key is `serialized_blocks`. With a `?? ''` fallback, every document hashed `sha256('')`. So the reported "`serializedBlocks` SHA-256 — 0 differing" compared 383 identical hashes of the empty string on both sides. It could not have detected a markup change.

The other fields in that fingerprint used correct keys and were genuinely compared: `blocks`, `diagnostics`, `fallbacks`, `assets`, asset keys and `coverage`, with the non-zero totals shown above. So the run was not worthless — but the single most important signal, the emitted block markup, was not actually checked.

### Re-run with the corrected harness

The harness now fails hard if `serialized_blocks` is absent rather than falling back to an empty string, and records markup length alongside the hash so a vacuous capture is self-evident.

Baseline captured on a clean `trunk` worktree at `24b115ac`, compared against this branch at `0f10c2cb`:

```
documents      : 383
markup bytes   : 22,047,755
distinct hashes: 383/383      (fingerprint is discriminating)
DIFFERING      : 0
threw          : 0
```

**The conclusion is unchanged — this refactor is behavior-preserving — but it is now actually evidenced.** The 383 distinct hashes over 22 MB of markup are what makes the comparison meaningful; previously that figure would have been 1.

CI remains green across PHP 8.2–8.5, and `composer test` covers the parity and contract suites independently of this harness.

### Why this got through, and what changes

The harness was validated for *reach* — I instrumented the nine entry points and confirmed they fire — but not for *discrimination*. Coverage of the code under change says nothing about whether the recorded signal can distinguish outputs. A fingerprint that is constant across every input passes any before/after comparison.

Carried into the remaining slices: assert the fingerprint varies across the corpus before trusting a zero-diff result. A single distinct hash across 383 documents is the tell.

---

*AI assistance disclosure: this correction was found and drafted by Claude Sonnet 4.6 running in Claude Code, operated by @chubes4. The defect surfaced while the AI was reading `TransformerResult::toArray()` for the next slice; it then rebuilt the harness, recaptured a clean-`trunk` baseline in a separate worktree, and re-ran the comparison. Reviewed by a human before posting.*

PR #1492, opened 2026-09-01
Keep the source's paint on a pushed-down link wrapper

gmovr:
## Problem

An `<a href>` that wraps block-level content becomes a `core/group`, and its link is pushed down onto the RichText blocks inside it as an inline `<a>` around their text (`HtmlTransformer::convertLinkWrapperGroup()` → `propagateInlineLink()`).

That rebuild puts the anchor **inside** the block that the source's painting element became. The anchor, not that element, is now the text's nearest painted ancestor, so the browser's default link colour replaces the colour the source painted.

Measured on `https://www.maneethical.com` (Wix), imported through Studio + static-site-importer, Playwright at 1280x900. The source menu item is:

```html
<a class="qMvpu5" href="/services/index.html">
  <div class="EWeavx"><div>
    <p class="wGxoBM" id="comp-k3mm4s7p2label">SERVICES</p>
  </div></div>
</a>
```

with `.wGxoBM{color:rgb(var(--txt,var(--color_15,color_15)))}` — the `<p>` inside the anchor is what paints the label.

The import emits `<p class="wGxoBM …"><a href="/services">SERVICES</a></p>`. The class still paints the `<p>` grey, but the rebuilt anchor repaints the text:

| label (front page) | source | import before | import after |
|---|---|---|---|
| HOME (current page) | `rgb(177, 211, 187)` | `rgb(0, 0, 238)` | `rgb(177, 211, 187)` |
| ABOUT | `rgb(65, 65, 65)` | `rgb(0, 0, 238)` | `rgb(65, 65, 65)` |
| SERVICES | `rgb(65, 65, 65)` | `rgb(0, 0, 238)` | `rgb(65, 65, 65)` |
| PRODUCTS | `rgb(65, 65, 65)` | `rgb(0, 0, 238)` | `rgb(65, 65, 65)` |
| CONTACT | `rgb(65, 65, 65)` | `rgb(0, 0, 238)` | `rgb(65, 65, 65)` |
| Shop | `rgb(65, 65, 65)` | `rgb(0, 0, 238)` | `rgb(65, 65, 65)` |

Text decoration was already carried across this rebuild (`textDecoration` in `linkPropagationAttributes()`); colour was not.

## Change

`linkPropagationAttributes()` now also asks whether an element **strictly inside** the wrapper paints every run of the wrapper's text, resolving the cascade winner via `specificityResolvedPresentationStyle()`. When it does, `propagateInlineLink()` marks the host block with `blocks-engine-propagated-link-color`, and `materializeAuthorStylesheet()` projects one rule after the author cascade:

```css
:root :where(.blocks-engine-propagated-link-color)>a{color:inherit}
```

Inheriting — rather than stamping a resolved value — restores whatever the source painted, because the block being inherited from still carries the source's own styling hooks. It also reproduces per-page state colours for free (the current item stays green on each page, see the table).

The condition is deliberately narrow, and nothing about it is site- or class-specific:

- **A colour resolved on the wrapper itself does not count.** It has no painting element inside the anchor to lose — source rules that reach the anchor by type still reach the rebuilt one — so no repaint is owed.
- **Text sitting directly in the wrapper does not count**, for the same reason; a wrapper holding any is left exactly as it is today.
- **Non-painting keywords** (`inherit`, `initial`, `unset`, `revert`, `currentColor`) do not count.
- **A wrapper whose text the source left to the browser's link colour is untouched**, so that parity is preserved by doing nothing.

The rule follows the author cascade so an authored rule can still repaint the link deliberately.

## Proof

Rebuilt the paired static-site-importer dev ZIP against this branch and re-ran a real 4-page import of the Wix capture, then re-measured with Playwright. Painted colour of every text run inside a link, source vs import, across `/`, `/about/`, `/services/`:

| | before | after |
|---|---|---|
| matching | 8 | 35 |
| differing | 30 | 3 |

The 3 remaining differences are one unrelated footer credit link (`Lumo Design Studio`), identical before and after — untouched by this change.

Import quality gates unchanged: 0 fallback blocks, 0 invalid blocks, 0 core/html, validation `passed`.

`tests/unit/link-wrapper-text-color-parity.php` locks in the emitted carrier, the projected rule, block validity, determinism, and each of the four negative cases above. `composer test` and `tests/contract/production-acceptance-matrix.php` pass locally (292 parity fixtures unchanged).

## Not in this PR

The same import also renders three source-hidden `More`-menu items and a screen-reader hint with real geometry instead of the source's zero geometry. That is a separate root cause (`StyleResolver::stripFrozenHiddenState()`) and is filed separately.

https://claude.ai/code/session_01MtVE4gr2RA7GE3FKbBS7C5

### DIFF
--- a/php-transformer/composer.json
+++ b/php-transformer/composer.json
@@ -189,7 +189,8 @@
             "php tests/unit/button-link-dispatcher.php",
             "php tests/unit/button-element-converter.php",
             "php tests/unit/runtime-resource-element-converter.php",
-            "php tests/unit/link-bearing-element-lowering.php"
+            "php tests/unit/link-bearing-element-lowering.php",
+            "php tests/unit/link-wrapper-text-color-parity.php"
     ],
     "test:parity": "php tests/parity/run.php",
     "test:migration:examples": "php tests/migration/examples.php",
--- a/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
+++ b/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
@@ -398,6 +398,15 @@ final class HtmlTransformer
      */
     public const EMPTY_RUNTIME_TARGET_CLASS = 'blocks-engine-empty-runtime-target';
 
+    /**
+     * Marks a RichText block whose text was painted by the source, and into
+     * whose content a content-wrapping anchor's link had to be pushed. The
+     * synthesized inline `<a>` becomes the text's nearest painted ancestor, so
+     * the browser's link colour would replace the source colour; the projected
+     * rule makes that anchor inherit its host block's colour instead.
+     */
+    private const PROPAGATED_LINK_COLOR_CARRIER_CLASS = 'blocks-engine-propagated-link-color';
+
     private const CSS_OWNED_LAYOUT_CLASS = 'blocks-engine-css-owned-layout';
 
     private const CSS_OWNED_FLOW_CLASS = 'blocks-engine-css-owned-flow';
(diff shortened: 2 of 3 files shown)


gmovr:
The first run of the solved-site promotion gate failed on `15-saas` with a 6-pixel mismatch (6 / 9,955,320 px = 0.00%, threshold 0.00%), attributed to a 2x2 region over `section#product > … > div.tag` ("How it works") — a mono-font badge, i.e. anti-aliasing noise, not a restyle.

Verified it is unrelated to this change: transforming `fixtures/solved/15-saas/index.html` with this branch and with `trunk` produces **byte-identical** output — same serialized blocks and same generated stylesheets — and the new carrier class fires zero times on that fixture.

Re-ran the gate; it passes. All checks are green.

chubes4:
Yes, the solved site gate can be a bit flaky sometimes, I haven't quite figured out what causes that occasional mismatch. Thanks for the PR!

Issue #1482, opened 2026-09-01
Model navigation once and project its unsupported residue generically


chubes4:
## Problem

Navigation recognition, native mapping, and re-presentation are spread across four files with no shared model, and every source idiom that core blocks cannot serialize costs four separate additions: a new marker vocabulary, a new registry field, a new CSS emission site, and a new projector method.

Current cost, reproducible from the repo root:

```sh
# 16 distinct marker vocabularies
grep -rhoE "blocks-engine-[a-z-]*navigation[a-z-]*|blocks-engine-navigation-[a-z-]*" php-transformer/src/ \
  | sed -E "s/-[0-9a-f]{8,}$//" | sort -u

# 3,592 lines across four navigation files
wc -l php-transformer/src/HtmlToBlocks/Patterns/NavigationPattern.php \
      php-transformer/src/HtmlToBlocks/Style/NavigationStyleProjector.php \
      php-transformer/src/HtmlToBlocks/NavigationBlockNormalizer.php \
      php-transformer/src/HtmlToBlocks/Patterns/NavigationPatternContext.php
```

The vocabularies today: `brand-navigation-carrier`, `current-navigation-item`, `current-navigation-underline`, `direct-navigation`, `direct-navigation-link-color-*`, `direct-navigation-reset-*`, `inline-navigation`, `list-navigation`, `native-navigation-toggle-*`, `native-responsive-navigation`, `navigation-current-color-*`, `navigation-link-color-*`, `navigation-link-color-states-*`, `navigation-link-icon-*`, `navigation-submenu-background-*`, `projected-dialog-navigation`.

### The failure mode this shape produces

Because the pattern and the projector each independently re-derive the same fact, they can disagree — and the projector then discards the styling silently.

`NavigationStyleProjector::directNavigationSupportCss()` recomputes a hash the pattern already computed, and drops the rule when the class is absent:

```php
$expectedMarker = 'blocks-engine-direct-navigation-link-color-' . substr(hash('sha256', $safeColor), 0, 12);
$classes = preg_split('/\s+/', trim((string) ($attrs['className'] ?? '')));
if ( ! in_array($expectedMarker, $classes, true) ) {
    continue;
}
```

`navigationLinkTextColorRules()` repeats the same shape with a second hash input (`$color . "\0" . $stateMask`). Any drift between the two derivations is an invisible loss of source presentation, which is a recurring source of "navigation still renders wrong" reports rather than a single bug.

## Expected contract

[...]

PR #1494, opened 2026-09-01
Recover navigation presentation the source inherits

chubes4:
Partially addresses #1482.

## Problem

A generated site's menu renders with none of the source's typography or colour. Measured against a public source at 1440px:

| | font-family | font-size | colour |
|---|---|---|---|
| Source | `helvetica-w01-roman` | 15.75px | `rgb(238, 255, 255)` |
| Generated | `Arial` | 10px | `rgb(0, 0, 0)` |

Those are the generated theme's body defaults, so nothing source-derived reaches the links.

The colour rule is generated and its stylesheet is enqueued, so this is not a delivery failure. It emits:

```css
… .blocks-engine-navigation-link-color-4881… > .wp-block-navigation-item__content { color: inherit }
```

`color: inherit`, because the source anchor's declared colour genuinely *is* `inherit`. Enumerating matching rules in both documents also finds no rule setting `font-family` or `font-size` on the anchor or its label span in either: the source's navigation typography is inherited too.

The projection is faithful to the source *declaration* and wrong about the *resolved value*. Copying `inherit` reproduces the source only while the ancestor that supplied the value is also reproduced, and native navigation replaces that structure.

## Change

Resolve what the source would have inherited by walking its ancestor chain, and state it on the navigation.

The recovered value is delivered as **CSS**, hung off classes the block already carries, rather than written onto the block. That placement is the substance of the change, not an implementation detail: an earlier attempt put the resolved value on `core/navigation`'s native attributes (`customTextColor`, `style.typography.*`) and broke `shared-shell-plan`. Shell identity compares block markup across documents, so a value that varies per page fragments one shared template part into one part per page. Keeping the value in CSS fixes the typography while leaving shell identity untouched.

Generated selectors use only the source's own classes; engine markers are excluded, so this adds no marker vocabulary.

## Verification

- 292 parity fixtures pass unchanged.
- `shared-shell-plan` passes, which is the contract the earlier attempt broke.
- New contract coverage asserts all three halves: the rule is recovered, it stays out of the navigation block's markup, and a menu that inherits nothing gets no fabricated rule.

## Scope

This addresses the inherited-presentation failure described in #1482. The broader model in that issue — one `NavigationModel`, a typed residue list, and a single emitter replacing the 16 marker vocabularies — remains open.

## AI assistance disclosure

Implemented with AI assistance: Claude Sonnet 4.5 running in the opencode CLI agent. The AI measured the computed styles and matching CSS rules in both documents, identified `color: inherit` as the emitted value, prototyped the native-attribute approach and diagnosed its shared-shell regression, implemented the CSS-delivered approach instead, and ran the suite. A human directed the work and the placement decision.

### DIFF
--- a/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
+++ b/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
@@ -2088,6 +2088,9 @@ final class HtmlTransformer
         foreach ( $this->navigationStyleProjector->navigationLinkTextColorRules($serializedBlocks) as $navigationLinkTextColorRule ) {
             $afterAuthorCssParts[] = $navigationLinkTextColorRule;
         }
+        foreach ( $this->generatedSupportStyles()->navigationInheritedPresentationRules() as $navigationInheritedRule ) {
+            $afterAuthorCssParts[] = $navigationInheritedRule;
+        }
         foreach ( $this->navigationStyleProjector->navigationLinkIconRules($serializedBlocks) as $navigationLinkIconRule ) {
             $afterAuthorCssParts[] = $navigationLinkIconRule;
         }
@@ -2580,7 +2583,10 @@ final class HtmlTransformer
                 fn (DOMElement $sourceElement): array => $this->navigationStyleProjector->navigationColorInteractionStates($sourceElement),
                 fn (DOMElement $sourceElement): string => $this->navigationToggleSuppressor->navigationOverlayMenu($sourceElement),
                 fn (DOMElement $sourceElement): string => $this->responsiveNavigationToggleMarker($sourceElement),
-                fn (DOMElement $sourceElement): string => $this->navigationLinkIconMarker($sourceElement)
+                fn (DOMElement $sourceElement): string => $this->navigationLinkIconMarker($sourceElement),
+                function (DOMElement $sourceElement, array $authorClasses): void {
+                    $this->recordInheritedNavigationPresentation($sourceElement, $authorClasses);
+                }
             ),
             new MediaPatternContext(
                 fn (DOMElement $sourceElement): string => $this->styleResolver->mergedPresentationStyle($sourceElement),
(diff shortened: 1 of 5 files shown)


chubes4:
Closing this unmerged: verifying it against a real generated site rather than only the contract fixture showed the premise is wrong.

The change resolves navigation typography by walking the source **ancestor** chain. On a real site that produces exactly the destination default it was meant to replace:

```css
.wp-block-navigation._root_s6hzk_2.navbar._navbar_cohan_1._ltr_cohan_24._container_s6hzk_173{
  font-family:Arial,Helvetica,sans-serif;font-size:10px
}
```

Walking the ancestors of the menu in the source shows why:

| element | computed font |
|---|---|
| `a` | `helvetica-w01-roman` 15.75px |
| `div.item._labelContainer_1sfgg_18` | `helvetica-w01-roman` 15.75px |
| `div._itemWrapper_1sfgg_10` | `Arial` 10px |
| `ul`, `nav`, and every ancestor above | `Arial` 10px |

The navigation's typography is declared on `._labelContainer_1sfgg_18`, a wrapper **inside** the menu, not on an ancestor of it. The `nav` itself genuinely computes to the document default, so an ancestor walk faithfully returns that default.

The earlier evidence that led to "inherited from an ancestor" was a rule enumeration finding no `font-family`/`font-size` rule matching the anchor or its label span. That observation is correct but the conclusion did not follow: the declaring rule matches an intermediate wrapper that is neither the anchor, its label span, nor an ancestor of the `nav`. (I also checked the obvious way that probe could lie — all 22 source stylesheets are same-origin and readable, so nothing was silently skipped.)

The corrected root cause is recorded on #1482. The shell-identity constraint this PR was designed around still holds and is worth keeping: recovered presentation must be delivered as CSS rather than written onto the block, or documents sharing a shell stop collapsing into one template part.

## AI assistance disclosure

Investigated with AI assistance: Claude Sonnet 4.5 running in the opencode CLI agent. The AI built and verified this change against a fresh generated site, found the emitted rule reproduced the destination default, walked the source ancestor chain to locate the true declaring element, and withdrew the change. A human directed the work.

PR #1496, opened 2026-09-01
Recover menu presentation from replaced source wrappers

chubes4:
Partially addresses #1482.

## Problem

A generated site's menu renders with none of the source's colour or type. Measured against a public source at 1440px:

| | font-family | font-size | colour |
|---|---|---|---|
| Source | `helvetica-w01-roman` | 15.75px | `rgb(238, 255, 255)` |
| Generated | `Arial` | 10px | `rgb(0, 0, 0)` |

Those are the generated theme's body defaults, so nothing source-derived reaches the links.

The cause is **structural replacement**, not a missing stylesheet. `core/navigation` emits its own item markup, and the source wrapper that carried the menu's presentation is not among the surviving elements. Walking the source shows the presentation appears on a wrapper *inside* the menu:

| element | computed font |
|---|---|
| `a` | `helvetica-w01-roman` 15.75px |
| `div.item._labelContainer_…` | `helvetica-w01-roman` 15.75px |
| `div._itemWrapper_…` | `Arial` 10px |
| `ul`, `nav`, every ancestor above | `Arial` 10px |

## Change

Recover presentation from the source element the native item stands in for, bounded by the navigation itself, and state it on the native counterpart. Delivery is **CSS**, hung off classes the block already carries.

That placement is the substance of the change. An earlier attempt wrote resolved values onto `core/navigation`'s native attributes and broke `shared-shell-plan`: shell identity compares block markup across documents, so a value that varies per page splits one shared template part into one part per page. A second attempt carried the replaced wrapper's classes onto the item and broke the same contract, because those wrappers can hold per-route state classes. Keeping recovery in CSS leaves shell identity untouched.

Also adds `StyleResolver::unsupportedSelectorDeclaration()`, which reads a declaration from a rule naming the element through a selector the matcher cannot evaluate (`:is`, `:where`, `:not`, `:has`).

## Verified on a generated site

Colour is recovered exactly:

| | before | after | source |
|---|---|---|---|
| link colour | `rgb(0, 0, 0)` | **`rgb(238, 255, 255)`** | `rgb(238, 255, 255)` |

Emitted rule:

```css
.wp-block-navigation._root_s6hzk_2.navbar… .wp-block-navigation-item__content{
  color:var(--wst-base-1-color);letter-spacing:0em;text-transform:none
}
```

## Known remaining gap

Font family and size are **not** yet recovered on that site, and this PR does not claim them. The menu's type is stated as:

```css
#main_MF :where(.comp-mb7ogqrp_r_comp-mqqn4p2c).menu .item { font: var(--wst-paragraph-2-font) }
```

Two properties of that rule defeat recovery: the selector uses `:where()`, which the matcher treats as unsupported, and the value is a `font` shorthand rather than longhands. The rule is present in the captured stylesheet, so nothing is lost at capture. Tracked separately with reproduction.

## Verification

- 292 parity fixtures pass unchanged.
- `shared-shell-plan` passes, which is the contract two earlier approaches broke.
- Contract coverage asserts recovery happens, reads the source menu rather than the surrounding document default, stays out of the navigation block's markup, and fabricates nothing for a menu with no distinct presentation.

## AI assistance disclosure

Implemented with AI assistance: Claude Sonnet 4.5 running in the opencode CLI agent. The AI measured both documents, used DevTools Protocol to find the winning rule, prototyped and withdrew two approaches that regressed shell identity, implemented this one, and verified each iteration against a freshly generated site. A human directed the work.

### DIFF
--- a/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
+++ b/php-transformer/src/HtmlToBlocks/HtmlTransformer.php
@@ -2088,6 +2088,9 @@ final class HtmlTransformer
         foreach ( $this->navigationStyleProjector->navigationLinkTextColorRules($serializedBlocks) as $navigationLinkTextColorRule ) {
             $afterAuthorCssParts[] = $navigationLinkTextColorRule;
         }
+        foreach ( $this->generatedSupportStyles()->navigationInheritedPresentationRules() as $navigationInheritedRule ) {
+            $afterAuthorCssParts[] = $navigationInheritedRule;
+        }
         foreach ( $this->navigationStyleProjector->navigationLinkIconRules($serializedBlocks) as $navigationLinkIconRule ) {
             $afterAuthorCssParts[] = $navigationLinkIconRule;
         }
@@ -2580,7 +2583,10 @@ final class HtmlTransformer
                 fn (DOMElement $sourceElement): array => $this->navigationStyleProjector->navigationColorInteractionStates($sourceElement),
                 fn (DOMElement $sourceElement): string => $this->navigationToggleSuppressor->navigationOverlayMenu($sourceElement),
                 fn (DOMElement $sourceElement): string => $this->responsiveNavigationToggleMarker($sourceElement),
-                fn (DOMElement $sourceElement): string => $this->navigationLinkIconMarker($sourceElement)
+                fn (DOMElement $sourceElement): string => $this->navigationLinkIconMarker($sourceElement),
+                function (DOMElement $sourceElement, array $authorClasses): void {
+                    $this->recordInheritedNavigationPresentation($sourceElement, $authorClasses);
+                }
             ),
             new MediaPatternContext(
                 fn (DOMElement $sourceElement): string => $this->styleResolver->mergedPresentationStyle($sourceElement),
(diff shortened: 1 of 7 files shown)


chubes4:
## Navigation type is now recovered — verified on a generated site

| | font-family | font-size | colour |
|---|---|---|---|
| Source | `helvetica-w01-roman` | 15.75px | `rgb(238, 255, 255)` |
| Generated | `helvetica-w01-roman` | 15.75px | `rgb(238, 255, 255)` |

Exact on all three. The "known remaining gap" in the description is closed, and #1497 turned out to be a misdiagnosis on my part — details below.

### What the gap actually was

`:is()`, `:where()` and `:not()` are already parsed and matched by `CssSelectorMatcher`, including the zero-specificity contribution for `:where()`. The selector was never the problem.

The menu's type is stated by this rule, which **is** captured and **is** enqueued on the front page:

```css
#main_MF :where(.comp-mb7ogqrp_r_comp-mqqn4p2c).menu .item { font: var(--wst-paragraph-2-font) }
```

Every part of that selector still resolves in the generated document — `#main_MF`, the `.comp-…` class, and `.menu` are all present — except the rightmost `.item`. That class lived on the wrapper between the list item and its anchor, which `core/navigation` replaces with its own item markup. The rule loads and matches nothing.

### The change

Carry the classes of those replaced wrappers onto the native item, so source rules written against them keep matching.

The constraint is that this touches block markup, and shell identity compares markup across documents. Carrying every wrapper class regressed `shared-shell-plan`, because a wrapper can hold route-current state that differs per page. Carrying only the classes **shared by every item in the menu** excludes that state by construction rather than by name matching, and keeps documents that share a shell byte-identical.

One unit assertion was tightened rather than satisfied. `navigation-nested-authored-anchor` asserted `! str_contains($markup, 'anchor-carrier')` as a proxy for "the carrier did not survive as an anonymous group or paragraph block". The change keeps that intent — no group or paragraph blocks appear inside the navigation — while deliberately carrying the class, so the assertion now tests the intent directly and additionally asserts the class is carried.

### Verification

- 292 parity fixtures pass unchanged.
- `shared-shell-plan` passes.
- Verified on a freshly generated site: navigation type and colour match the source exactly, footer social icons still render at 23×23 from their recovered artwork, and no other measured property regressed.

### AI assistance disclosure

Investigated and implemented with AI assistance: Claude Sonnet 4.5 running in the opencode CLI agent. The AI used DevTools Protocol to identify the winning rule, confirmed the rule was captured and enqueued, isolated the missing `.item` class as the sole reason it stopped matching, and verified each iteration against a freshly generated site. A human directed the work.

chubes4:
## Also fixes duplicated menu painting

A source menu with a painted region rendered that region **twice**, stacked on itself.

WordPress copies a navigation block's classes onto both the `nav` and its responsive container:

```
NAV: … _root_s6hzk_2 navbar … _container_s6hzk_173 …
UL : wp-block-navigation__container … _root_s6hzk_2 navbar … _container_s6hzk_173 …
```

So any source rule painting the menu through one of those classes matches both elements.

Measured in the header, before:

| | painted / rounded elements |
|---|---|
| Source | `NAV` [298×30] radius 112.5px — one |
| Generated | `NAV` [298×38] radius 112.5px **and** `UL.wp-block-navigation__container` [289×29] radius 112.5px — two |

After: the generated document paints once, matching the source.

Where the source paints the menu, the paint is neutralised on the inner container so it is stated once. The reset is gated on the source actually painting, so a menu that paints nothing gets no rule.

One detail worth noting for review: the reset uses a descendant combinator rather than a child one, because core nests the container inside `.wp-block-navigation__responsive-container` — a child combinator emits a valid-looking rule that never matches.

### Verification on a freshly generated site

- Menu paints once, matching the source.
- Navigation type and colour unchanged and still exact: `helvetica-w01-roman` 15.75px `rgb(238,255,255)`.
- Footer social icons still render at 23×23 from recovered artwork.
- 292 parity fixtures pass; `shared-shell-plan` passes.

### AI assistance disclosure

Implemented with AI assistance: Claude Sonnet 4.5 running in the opencode CLI agent. The AI measured painted regions in both documents, found the duplicated class application, implemented the gated reset, and verified against a freshly generated site. A human directed the work.

PR #1631, opened 2026-09-09
fix(php-transformer): honor explicit block flow before inferring columns

### LINKED ISSUE #1628: Preserve injected responsive document roots as block shells
## Root Cause

Blocks Engine PHP transformer's generic `ColumnsPattern` promoted a normal-flow container to `core/columns` from its sidebar/content child semantics without honoring explicit `display:block` evidence. The resulting core columns forced the source-flow children into equal-width flex columns.

The responsive document wrapper is merely the triggering input. No data-liberation marker special case is required. The generic recognizer now declines column inference for explicit block, inline, flow-root, or contents display modes, while preserving semantic inference when no authored display evidence exists.

## Verification

- Plain-class two-child `aside` + `main` HTML with `display:block` now emits `core/group`, not `core/columns`.
- The transformer shared static geometry parity harness reports a 1.0 score for the block-flow case including a hidden lightbox sibling.
- Full PHP transformer test suite passes.

## AI assistance disclosure

GPT-5.6 Terra via OpenCode, orchestrated by GPT-6 Astra via OpenCode, traced the generic transformer selection path, reproduced the issue, implemented the regression test and fix, and ran verification.

chubes4:
## Summary

Fixes #1628.

Honor explicit normal-flow display evidence before inferring native Columns from semantic sidebar/content children. A wrapper with authored `display:block` must remain a Group rather than acquiring an equal-width flex row and a Gutenberg column gap.

The reduced fixture uses ordinary classes and semantic children. The fix adds no source-platform marker or site-specific exception.

## Verification

- `php php-transformer/tests/unit/block-style-support-conversion.php`: 138 assertions passed.
- `composer test` from `php-transformer/`: canonical, unit, 303 parity fixtures, distribution-shape and package-install checks passed.
- Shared static-style geometry comparison of the reduced normal-flow fixture reports geometry, property parity and coverage of 1.
- PHP syntax and diff checks passed.

This fixes one proven layout inference defect. Full Potted site parity remains an independent acceptance gate after refreshing the paired importer package and reimporting.

## AI Assistance

GPT-5.6 Terra via OpenCode traced the source and generated geometry, corrected the initial ownership diagnosis, implemented the generic inference guard and verified it. GPT-6 Astra via OpenCode reviewed the scope and diff and prepared publication. Chris directed the work and authorized direct execution while the orchestration control plane was blocked.

### DIFF
--- a/php-transformer/src/HtmlToBlocks/Patterns/ColumnsPattern.php
+++ b/php-transformer/src/HtmlToBlocks/Patterns/ColumnsPattern.php
@@ -157,6 +157,12 @@ final class ColumnsPattern implements PatternRecognizerInterface
             return false;
         }
 
+        // Semantic sidebar/content names can suggest a split layout only when
+        // author CSS does not explicitly retain normal document flow.
+        if ( preg_match('/(?:^|;)\s*display\s*:\s*(?:block|inline|flow-root|contents)\b/', $style) ) {
+            return false;
+        }
+
         // core/columns is a flex layout. Preserve resolved grid containers as
         // groups so their source classes continue to control track geometry.
         if ( preg_match('/(?:^|;)\s*display\s*:\s*(?:inline-)?grid\b/', $style) ) {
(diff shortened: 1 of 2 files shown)


chubes4:
Rebased onto current trunk and re-verified: `composer test` green locally (canonical, parity, packaging).

The previous red CI was infrastructure, not code — the Generic visual parity tools job failed during `apt` package installation ("Some index files failed to download", exit 100) on 2026-09-09, and the composer suites passed in that same run.

AI assistance: OpenAI GPT-6 Astra via OpenCode rebased and re-verified this branch under Chris Huber's direction.

PR #1716, opened 2026-09-14
Fix selector-cache churn on large imports

chubes4:
## Summary

Raises the bounded per-document selector match cache to retain repeated CSS selector lookups on large source DOMs.

## Evidence

The retained Jen Derrick `website/index.html` page plan (1.07MB HTML, 1.75MB CSS) produced 247,302 selector-match executions and 239,110 evictions with the prior 4,096-entry cache. The new 131,072-entry bound reduced those to 180,858 executions and 22,483 evictions, with a measured 198MB peak process memory.

The first no-profiler baseline was 34.6s. Candidate measurements were 22.7s and 41.6s on a variable local host, so this PR does not claim a stable wall-clock improvement until CI or a controlled machine confirms it.

## Testing

- `php tests/unit/css-selector-matcher.php`
- `php tests/contract/staged-artifact-compilation.php`
- `composer test` (reaches an existing failure in `tests/unit/rich-text-svg-transaction.php`; the failure reproduces in isolation and is unrelated to these files)

## AI assistance

OpenAI gpt-5.6-terra via OpenCode investigated the retained workload, profiled selector-cache churn, implemented the bounded-cache adjustment, and ran the listed verification commands.


chubes4:
Rebased onto trunk and fixed the red CI, which was a real failure rather than a flake.

Raising `MAX_MATCHES` to 131,072 left `tests/unit/html-transformer-shared-analysis-cache.php` asserting numbers derived from the old 4,096 bound, so every `composer test` job failed on `Read-only conversion retains selector and structural declaration results across repeated passes`.

Rather than restate the constant in both tests, the bound is now injectable per instance and defaults to `MAX_MATCHES`:

- eviction is bound *behaviour*, so the selector-matcher unit test now proves it at a capacity of 8 instead of filling 131,072 entries — same coverage, and that loop no longer dominates the job;
- a zero capacity is refused rather than evicting every result it just computed;
- the default bound is asserted to retain what a small document needs;
- the shared-analysis pressure case keeps its 4,097-element document and now records the measured result under the new bound: 4,098 misses, 1 hit, **0 evictions**, 4,098 peak entries, with structural declaration builds/hits unchanged at 4,099/8,201. Those numbers were measured, not predicted.

Zero evictions on a 4,097-element document is the change working: a large source stops discarding hot selector matches only to recompute them.

`composer test` green locally across canonical, parity and packaging.

AI assistance: OpenAI GPT-6 Astra via OpenCode diagnosed the failure, made the bound injectable, and re-measured the pressure metrics under Chris Huber's direction.

PR #1804, opened 2026-09-16
fix(php-transformer): find scroll-states.json under website/

chubes4:
## Summary

Studio stages Data Liberation captures from the website root, so \`scroll-states.json\` arrives as \`website/scroll-states.json\`. The projector only looked for the exact basename at the artifact root and dropped per-page header scroll evidence.

Match those sidecars by basename.

## Test plan

- [x] \`php php-transformer/tests/unit/scroll-state-projector.php\`
- [ ] Annie recapture: Research header transparent at rest, solid on scroll

AI assistance: grok-4.6 via OpenCode. Used to trace missing scroll-state projection from capture sidecars into the compiler file list.


chubes4:
Closing: Blocks Engine should not know Data Liberation stages captures under website/. The compiler contract stays `scroll-states.json` at the artifact root. Studio/SSI should pass that sidecar at the path the compiler already expects.

AI assistance: grok-4.6 via OpenCode.

PR #1889, opened 2026-09-16
fix(php-transformer): replay nav anchor selectors inside conditional groups

chubes4:
## Problem

Source nav anchor selectors are replayed against `core/navigation`'s wrapper markup so `.jumpnav a` still reaches `.wp-block-navigation-item__content`. That replay only walked **unconditional** rules.

A responsive menu states its compact anchor box inside a breakpoint:

```css
.jumpnav a { padding: 0.35rem 0.85rem }
@media (max-width: 640px) { .jumpnav a { padding: 0.35rem 0.5rem } }
```

Only the first rule was replayed, so the menu kept its widest padding at every width. `@supports`, `@layer`, `@container`, and `@scope` groups were dropped the same way, and nested groups entirely.

## Fix

Recurse into conditional groups and retain their nesting, matching how `responsiveRootCompatCssRules()` already handles the same shape:

```css
@media (max-width: 640px) {.jumpnav.wp-block-navigation .wp-block-navigation-item__content { padding:0.35rem 0.5rem }}
```

A group whose rules map to no nav anchor emits nothing rather than an empty group.

## Validation

- `php php-transformer/tests/unit/navigation-anchor-media-replay.php` — 10 assertions, registered in `composer test:unit`. 8 fail on trunk.
- `composer test:unit` — full suite, exit 0.
- `composer test:parity` — 311 fixtures pass.
- `composer validate --strict`

## AI assistance

- **Model:** grok-4.6
- **Tool:** OpenCode
- **Used for:** Compared imported menu link boxes against source at two viewports, traced the missing compact padding to the unconditional-only replay, and recursed the pass.


chubes4:
Closing: the approach is wrong.

Recursing the anchor replay into conditional groups also replays selectors the structure pass has not rewritten, so a descendant list selector maps to nonsense:

```css
/* from: body.menu-ready .desktop-nav ul.site-menu>li a { … } */
body.menu-ready .desktop-nav ul.site-menu>li.wp-block-navigation .wp-block-navigation-item__content { … }
```

`.wp-block-navigation` lands on the `li` instead of the menu, which is why `artifact navigation projection replaces non-serialized source list markers with core navigation item selectors` fails on 8.2–8.5.

The conditional anchor rules need the same structure rewrite the unconditional path gets, ordered after it — not a bare recursion of the anchor mapper. Reopening with that once the ordering is right.

## AI assistance
- **Model:** grok-4.6
- **Tool:** OpenCode
- **Used for:** Reproduced the canonical contract failure locally and diagnosed the bad selector mapping.

PR #1939, opened 2026-09-17
fix(php-transformer): keep the band a collapsed menu item reserved

chubes4:
## Summary
`core/navigation` takes navigation-link children, so the wrapper chain a source menu item sat inside cannot survive conversion. In a mesh-authored menu that chain is where the item's vertical placement lives: a repeater stamps one template per row, and the designer's position inside the row cell compiles to block-axis margins on those wrappers.

Dropping them does not cost padding, it costs the item's y coordinate — the menu jumps to the top of its landmark.

## Evidence
A Wix export whose menu is a repeater (`comp-k1tu9hbb__item-…`, one template per row). Measured on the rendered page at 1440px:

| | source | imported |
|---|---:|---:|
| menu row height | 94px | 30px |
| menu band height | 100px | 36px |
| first link `y` | 31px | **0px** |

The missing 64px is exactly the collapsed chain: `margin: 3px 0` on the gutter wrapper and `margin: 28px 0 30px` on the row cell. `3 + 28 = 31` above, `30 + 3 = 33` below, and `31 + 36 + 33 = 100` — the authored band.

`core/navigation` supports `spacing.blockGap` and no padding, so the band cannot ride on the block's own attributes; setting it there is silently dropped. It is projected onto the rendered host instead, behind a marker naming the band it carries, so two menus that reserved the same band share one rule and a menu that reserved none is never reached.

Run against that capture's real header and stylesheets:

```
NAV className: Exmq9 blocks-engine-navigation-band-t31-b33 …
RULE: .wp-block-navigation.blocks-engine-navigation-band-t31-b33{padding-top:31px;padding-bottom:33px}
```

## Why requiring the rows to agree is safe rather than lossy
A repeater stamps its rows from one template, so their wrappers are identical by construction. Demanding agreement therefore costs nothing on the case this targets, while refusing menus whose per-item margins mean something individual.

## Test plan
- [x] `php tests/unit/navigation-collapsed-item-band.php`, added here and registered in `test:canonical`. Covers the banded repeater, and pins the four refusals: inline-axis-only margins, rows that disagree, a band that is not a resolved pixel length, and a directly authored anchor list that never had a chain to collapse.
- [x] `composer run test:canonical`
- [x] Replayed against the capture's real `SITE_HEADER` plus its inline and linked stylesheets, output above.

## Note
The end-to-end re-import is not included: the development-package build is currently refused by a filesystem reserve policy on this machine, not by anything in this change. The transformer-level replay above uses the same source markup and CSS the import would consume.

## AI assistance
Grok 4.6 via OpenCode. Measured the menu's computed geometry against the live source to separate a spacing difference from a lost coordinate, traced the 64px to the specific wrappers the collapse discards, confirmed `core/navigation` does not support padding before choosing projection over block attributes, and replayed the real captured header to check the emitted band.


chubes4:
Moving this to draft: the change is correct where it is tested, but it does not yet reach a real import, so it would not fix the reported symptom.

**What holds.** Through `HtmlTransformer` the band is computed and projected exactly as intended, on the synthetic fixtures and on the capture's real `SITE_HEADER` with its real stylesheets:

```
NAV className: Exmq9 blocks-engine-navigation-band-t31-b33 …
RULE: .wp-block-navigation.blocks-engine-navigation-band-t31-b33{padding-top:31px;padding-bottom:33px}
```

**What does not.** Compiling the same capture through `ArtifactCompiler` — the path a real import takes — the marker never appears:

```
parts with band marker: 0 / 2
rule in assets: NO
```

and it is already absent from the compiled page markup, before any shell extraction:

```
index.html                          hasNav=Y marker=N
goldman-sachs-design-system/…       hasNav=Y marker=N
… (8 documents, all marker=N)
```

So the difference is in how the compiler drives the transform, not in shell extraction or the site plan. Re-importing confirms it end to end: the menu is still at `y=3` against the source's `y=31`, and the rendered navigation reports `padding-top: 0px`.

The likely cause is that the compiler normalizes the artifact before transforming, which changes either the wrapper chain the band is summed from or the resolved style available for it. That wants isolating against the normalized document rather than guessing, and it should land in the same PR as the fix so the two are reviewed together.

Not merging until the compiler path reproduces the transformer result.

## AI assistance
Grok 4.6 via OpenCode. Verified the fix at transformer level, then reproduced the import through `ArtifactCompiler` and bisected the loss to the compiled page markup.

PR #1984, opened 2026-09-18
fix(php-transformer): keep an anchor's decorative svg icon inside the link

chubes4:
## Problem

A text-bearing anchor's inline icon was hoisted out of the anchor into a separate icon-only anchor during conversion, breaking the link's inline layout and adding vertical height.

Source shape:
```html
<a class="inline-flex items-center gap-2" href="/renovations">LEARN MORE <svg width="14" height="14">…</svg></a>
```

Measured on the imported site vs the captured source artifact (homepage, 1280x900):

| | capture | WordPress before fix |
|---|---:|---:|
| anchors with BOTH text and an icon | 7 | **0** |
| icon-only anchors | 3 | **10** |

Every text+icon anchor was split; 3+7=10. Direct measurement of one "LEARN MORE" link:

```
CAPTURE   link {h:20, w:113, display:"inline-flex"}   icon {tag:"svg", 14x14, sameLine:true}
WORDPRESS link {h:17, w: 95, display:"inline"}        icon: null
```

I searched Automattic/blocks-engine and Automattic/static-site-importer for an existing issue and found none tracking this specific defect.

## Root cause

`ButtonLinkDispatcher::convertAnchor` decides whether an anchor's content needs a native group host (`convertLinkWrapperGroup`) using `SourceElementClassifier::hasBlockContentChildren()`, which flags **any** non-text-formatting child as block content — including `svg`, since `isInlineContentElement('svg')` is `false` (svg gets a different materialization path than typographic tags like `<em>`/`<span>`). That false positive sent a plain "text + decorative svg icon" anchor down the link-wrapper-group path, which converts each child independently and severs the icon from the anchor's text run — it lands in its own icon-only anchor.

## Fix

Extended `hasBlockContentChildren()` with an opt-in `$treatSvgAsInline` parameter. The button/link dispatcher passes `true` for its anchor-content decision, so a decorative `svg` beside text no longer forces block promotion. Every other existing caller (`dd`/list-item/paragraph block-content checks, ~9 call sites) keeps the current default (`false`) and is unaffected.

This mirrors an idiom the codebase already uses ad hoc in `SvgMaterializer::svgNeedsPhrasingHost()` (`'svg' !== $tag && ! isInlineContentElement($tag)`) — "svg counts as inline for this one structural decision" — rather than widening the shared `isInlineContentElement()`/`isInlineSourceElement()` predicates, which are used in 20+ places with different semantics (rich-text materialization, quote patterns, richText markers) where svg's distinct handling is intentional and shouldn't change.

**A prior session in this worktree had a partial, uncommitted fix** that swapped the dispatcher's check to the existing `hasFlowContentChildren()` predicate instead. I verified that predicate too broadly excludes `video`/`audio`/`canvas`/`select`/`textarea`/`picture` (everything in `isInlineSourceElement`'s list, not just `svg`) from "block content." Running the full parity suite with that version surfaced a real regression: the `html-whole-element-link-dropped-finding` fixture (`<a href="/watch"><video>…</video></a>` — a link wrapping a video with no navigable inner block) stopped emitting `core/group > core/video` with a dropped-link finding, and instead produced `core/buttons` with the video's raw HTML crammed into the button's text attribute (caught by a downstream fallback-reduction pass after the dispatcher fell back to `core/html`). I kept the prior session's diagnosis (the `isInlineContentElement('svg') → false` asymmetry) and its test additions, but replaced the fix itself with the narrower, svg-specific opt-in described above, and added a fixture-shaped regression test (`tests/unit/inline-icon-anchor.php`, test #11–13) locking in that the video case still promotes to a group.

## Verification

- `composer test:unit` — all unit suites pass (421 assertions across all files), including the new `tests/unit/inline-icon-anchor.php` (13 assertions).
- `composer test:canonical` — passes.
- `composer test:parity` — all 319 fixtures pass, including `html-whole-element-link-dropped-finding` (the one the prior session's approach regressed).
- `composer test:packaging` — passes.
- **Real render, end-to-end**: built a dev package from this branch (`node tools/build-dev-package.mjs --blocks-engine-path <this worktree> --blocks-engine-ref HEAD`) and re-imported the captured site (`aromatic-monolith-build-atelier.base44.app`) into a fresh WordPress instance via the Studio CLI.
  - Homepage "LEARN MORE" anchor now renders `<a class="inline-flex items-center gap-1 …" href="…">Learn More </a>` — one anchor, icon inside, measured `{w:113, h:20, display:"inline-flex"}`, matching the capture's `{h:20, w:113, display:"inline-flex"}` exactly.
  - Homepage anchor-count arithmetic now matches the capture exactly: 7 text+icon anchors, 3 icon-only anchors (previously 0 and 10).
  - "What We Build" 6-card grid: each card is now 408px (was reportedly 436px before the fix; capture is 412px — within normal cross-renderer tolerance, no longer +24px/card).
  - Page heights (capture → WordPress, 1280×900, before → after):
    - home: 4526 → was +48px over → now 4518 (−8px, small residual unrelated to this bug)
    - about: 1938 → 1938 (**exact**)
    - contact: 1487 → 1701 (+214px — **not** caused by this bug: the contact page has zero text+icon anchors in either capture or WordPress; the delta is a form-rendering difference, confirmed via DOM inspection, unrelated and out of scope)
    - labour: 1891 → 1891 (**exact**)
    - process: 2481 → 2481 (**exact**)
    - renovations: 4581 → 4581 (**exact**, 6 text+icon anchors correctly matched)
    - safety: 1987 → 1987 (**exact**)
    - work: 2465 → was +18px over → now 2483 (+18px — **not** caused by this bug either: zero text+icon anchors on this page in either capture or WordPress; pre-existing, unrelated)

So this fix takes the site from a reported 5/8-pixel-exact baseline to **6/8 pixel-exact pages** (about, labour, process, renovations, safety pixel-exact; home within 8px of exact and no longer has the +48px icon-hoisting defect). The remaining work (+18px) and contact (+214px) deltas are verified to be pre-existing, unrelated issues — neither page has any text+icon anchors, so this code path never executes for them.

## Reviewer verification steps

1. `cd php-transformer && composer test:unit && composer test:parity`
2. `php tests/unit/inline-icon-anchor.php` for the focused fixture-shaped test.
3. Diff review: `SourceElementClassifier::hasBlockContentChildren()` gains one optional bool param with a documented default (`false`) that preserves every other caller's behavior; the only behavioral call-site change is `ButtonLinkDispatcher::convertAnchor()` passing `true`.

## AI disclosure

Parts of this PR's diagnosis and diff were found already written and uncommitted in this worktree from a prior automated session (model: anthropic/claude-sonnet-5, tool: opencode) before I started; I reviewed that partial work, verified it against the real bug and the parity suite, found it introduced a regression (detailed above), and replaced the fix itself while keeping its correct diagnosis and expanding its test coverage. All work in this PR, including the fix approach used in the final diff, the regression-catching investigation, the new tests, and the end-to-end verification, was produced with AI assistance (model: anthropic/claude-sonnet-5, tool: opencode) under human supervision.


chubes4:
Closing as superseded/duplicate: while I was investigating and verifying this fix, #1979 (`fix(php-transformer): keep phrasing text-and-icon links as one anchor`) merged to trunk fixing the same root cause (svg being misclassified as block content in `ButtonLinkDispatcher`'s anchor-content decision). I independently reached the same diagnosis and verified #1979's merged fix resolves the exact reported bug end-to-end (single-anchor output, correct `{w:113,h:20,display:inline-flex}` geometry, all 320 parity fixtures green on trunk), so this PR adds no further value and would only create redundant/conflicting code. See the handoff log for full verification detail from my independent investigation, including a regression I found and avoided in an earlier draft of this fix (an anchor-wrapped `<video>` case).

Issue #2034, opened 2026-09-18
Residual vertical rhythm deltas after #2030: /blog/ +190 and captioned-figure pages -146


chubes4:
## Summary

#2030 fixed homepage post-list rhythm (`+248` -> `+31`) and improved four article pages (`+47` -> `+30` each). Its PR reported `/blog/` and `/hyundai-i30n/` as **unchanged**, and its "after" column was a prediction rather than a measurement.

Re-measured after merge, both moved — slightly worse:

| route | before #2030 | after #2030 | change |
| --- | ---: | ---: | ---: |
| `/` | +248 | **+31** | improved |
| 4 article pages | +47 | **+30** | improved |
| `/contact/` | 0 | 0 | exact |
| `/now/` | 0 | 0 | exact |
| `/blog/` | +175 | **+190** | **+15 worse** |
| `/hyundai-i30n/` | −129 | **−146** | **−17 worse** |

Measured at Blocks Engine `17b6fb8d`, Static Site Importer `945c3481`, DLA `03bcd668`, Playwright 1280x900 after lazy-load settle. Image, heading and `pre`/`code` counts match live on every route; no broken images.

#2030 is a clear net win and this is not a request to revert it. These are the two residual deltas it did not cover, now isolated.

## 1. `/blog/` — +190

#2030's author described this as "a structural list, not this path". It is the archive index: a year heading followed by dated post rows with excerpt and a "Read more" link. Superficially similar to the homepage list that #2030 fixed, but evidently reaching a different code path.

Worth establishing why the homepage fix does not apply here before writing anything — if the two lists differ only incidentally, the right fix may be making them share one path rather than adding a second.

## 2. `/hyundai-i30n/` — −146 (imported is **shorter** than source)

This one is self-inflicted and is a genuine design tension, not a simple bug.

The source markup is `<figure><p><img></p><figcaption>`. Blocks-engine #2018 correctly converts that to `core/image` with a caption — which necessarily drops the inner `<p>` wrapper, and the paragraph's margins go with it. Eight such figures on this page, and the page ends up 146px shorter than the source.

So the output is semantically correct and the geometry is wrong. Options worth weighing rather than assuming:

- Carry the dropped wrapper's representable vertical spacing onto the emitted `core/image`, the way other promotion paths carry wrapper presentation.
- Accept the deviation as correct-by-design and exclude it from parity expectations, documenting why.

[...]

PR #2039, opened 2026-09-18
test(php-transformer): expect synthetic paragraph on collapsed text wrappers

chubes4:
## Summary

Follow-up to #2036. That change marks childless visual text wrappers with `blocks-engine-synthetic-paragraph` so user-agent paragraph margins are neutralized. Three save-shape assertions in `block-style-support-conversion.php` still expected the old class list and failed Composer CI on every PHP version.

Update those assertions to include the synthetic marker. No production change.

## AI assistance

Produced with model `xai/grok-4.6` via `opencode` while iterating on #2036 CI.


chubes4:
Closing as obsolete — this is a test fix for a change that is no longer on `trunk`.

Sequence:

| commit | |
| --- | --- |
| `b42bc1f8` | #2036 merged — marks collapsed text wrappers `blocks-engine-synthetic-paragraph` |
| `7aca2378` | #2040 |
| `6e73ecba` | **#2041 reverts #2036** |

This PR updates three assertions to expect `blocks-engine-synthetic-paragraph` (7 occurrences in the diff), but the production change that emitted that marker is reverted. `trunk` is currently green at `6e73ecba`; merging this would assert a class the code no longer produces and take it red again.

The cosmetic half this addressed was only ever one of two failures in #2036. The other — a leaf that resolved its own `text-align` losing it to the wrapper, caught by the `html-box-chrome-wrapper-text-alignment` parity fixture — is the real blocker and is unaddressed here.

Both halves are now tracked in #2034, which covers the residual `/blog/` rhythm this was ultimately in service of. Whatever replaces #2036 should land its production change and its test expectations together, with that fixture green.

AI disclosure: reviewed and closed by Claude (Anthropic) in an agentic coding session, after verifying the revert on `trunk`, confirming the marker is absent from `src`, and checking current `trunk` CI is green. Directed by a human.

PR #2045, opened 2026-09-18
test(php-transformer): restore expectations stranded by the #2036 revert

chubes4:
## Trunk is currently red

`origin/trunk` has been failing since #2041 merged. Verified on a clean checkout with no other
changes:

```
FAIL: 26: typography-only card tier label collapses to a styled paragraph…
FAIL: 28: pure-text card result row carries its box chrome on one paragraph
FAIL: 29b: a pure-text eyebrow under a universal zero reset collapses…
Block style support conversion tests: 3 failed, 135 passed

Fixture html-box-chrome-wrapper-text-alignment failed expectation at serialized_blocks.
```

## Cause: an incomplete revert

#2036 changed behavior **and** updated the expectations that described it. #2041 reverted the
source behavior but left those expectations in place, so they now describe a shape trunk no longer
emits:

- three assertions in `tests/unit/block-style-support-conversion.php` still expect childless visual
  text wrappers to carry `blocks-engine-synthetic-paragraph`
- `tests/fixtures/parity/html-box-chrome-wrapper-text-alignment.json` still expects an
  `blocks-engine-inline-layout-carrier` paragraph with an inline `style="text-align:right"`

## Fix

Restores all four to their pre-#2036 form, which is exactly what trunk emits today. **No source
changes** — this is expectations-only.

Confirmed the restored fixture expectation matches trunk's real output:
```
<p class="has-text-align-right blocks-engine-synthetic-paragraph">
```

## Verification

```
php tests/unit/block-style-support-conversion.php   ->  138 passed
php tests/parity/run.php                            ->  338 fixture(s) passed
composer test                                       ->  full suite green
```

## Note on #2039

#2039 targets the same red, but it was authored against a tree that still contained #2036 and is now
stale: it expects `blocks-engine-inline-layout-carrier` where trunk emits
`blocks-engine-synthetic-paragraph`, and it conflicts with trunk. I rebased it locally to check and
its expectations still fail against current trunk. This PR supersedes it; #2039 should be closed.

## AI assistance disclosure

Model: Claude Sonnet 4.5. Tool: opencode. The AI found trunk red while verifying an unrelated PR,
reproduced the failures on a clean trunk worktree to rule out its own changes, bisected the cause to
the incomplete #2041 revert by diffing the expectations against `b42bc1f8^`, confirmed the pre-#2036
expectations match trunk's current output, applied the restoration and ran the suites. A human
directed the work.


chubes4:
Closing — superseded by #2042, which fixed this correctly.

Trunk was genuinely red at `6e73ecba` (reproduced on a clean worktree; CI agrees — that commit's PHP Transformer run concluded `failure`). The cause was as described here: #2041 reverted #2036's source behavior while leaving its updated expectations in place.

But the right remedy was to **reapply the source**, not to strip the expectations. #2042 (`b1d2ab7f`) did exactly that and trunk is green again. With #2036 reapplied, the three `blocks-engine-synthetic-paragraph` assertions and the `html-box-chrome-wrapper-text-alignment` fixture are correct as they stand, and this PR would re-break them.

Verified before closing: this branch's own CI output now contains `<p class="tier-name blocks-engine-synthetic-paragraph">Team</p>`, i.e. the marker these assertions expect.

No action needed. Also note #2039 remains stale against current trunk for the same underlying reason and is worth closing separately.

AI assistance disclosure: model Claude Sonnet 4.5 via opencode. The AI opened this PR after reproducing trunk's failure, then re-checked trunk's CI history before merging, found #2042 had landed the correct fix, and closed this rather than shipping a regression. A human directed the work.

PR #2046, opened 2026-09-18
fix(php-transformer): restore leaf alignment expectations after the #2036 revert

chubes4:
## Why

`trunk` is red on the **PHP Transformer** workflow at `6e73ecba`. The engine is correct; the test expectations are not.

#2040 adapted two sets of expectations to the behaviour #2036 introduced:

- the `html-box-chrome-wrapper-text-alignment` parity fixture
- three assertions in `tests/unit/block-style-support-conversion.php` (26, 28, 29b)

#2036 was then reverted in #2041, but #2040's expectation changes stayed. Nothing reverted them, so trunk fails on expectations that describe a behaviour no longer present.

## What this restores

`#2040` weakened the fixture to accept exactly the regression #2041 reverted:

| | #2040 expectation | restored |
|---|---|---|
| leaf block | `blocks-engine-inline-layout-carrier`, `align: center` | `blocks-engine-synthetic-paragraph`, `align: right` |
| serialized | `style="text-align:right"` | `<p class="has-text-align-right blocks-engine-synthetic-paragraph">` |

and flipped three assertions to a `blocks-engine-synthetic-paragraph` class that only appears with #2036 applied:

- `26` `<p class="tier-name">Team</p>`
- `28` `<p class="use-case-result">Launch faster</p>`
- `29b` `<p class="eyebrow">The Shop</p>`

The restored values are the pre-#2040 ones and match what the engine emits today. They also encode the fixture's stated intent — *"a leaf that resolved its own alignment keeps it"* — which #2041's revert description calls out as original intent rather than something #2036 should have rewritten.

**No source changes.** Test expectations only.

## Verified locally

| | before | after |
|---|---|---|
| `composer test` exit | 1 | **0** |
| `tests/unit/block-style-support-conversion.php` | 3 failed, 135 passed | **138 passed** |
| `tests/parity/run.php` | `html-box-chrome-wrapper-text-alignment` failed | **338 fixtures passed** |

138/338 matches the green baseline #2041 recorded for the revert.

## Note on #2039

#2039 moves in the opposite direction — it adds `blocks-engine-synthetic-paragraph` expectations. That was the right call while #2036 was on trunk; after #2041 it would re-break these three assertions. It should be closed in favour of this PR.

## AI assistance disclosure

Diagnosed and prepared with **Claude Sonnet 4.5** via **opencode**, running an orchestrated Data Liberation import loop. The model bisected the red workflow to #2040's expectation edits by running `tests/parity/run.php` and `tests/unit/block-style-support-conversion.php` against pristine `6e73ecba` versus the candidate, diffed `f99cc3dc..7aca2378` to recover the pre-#2040 values, and verified `composer test` exits 0. A human reviewed the change before publishing.


chubes4:
Obsolete — superseded by #2042.

I diagnosed this against trunk tip `6e73ecba` (the #2041 revert), where #2040's expectations no longer matched engine behaviour and `PHP Transformer` was failing. The correct resolution was to bring the behaviour back rather than walk the expectations back, which is what #2042 did by reapplying #2036.

Trunk is green at `b1d2ab7f` (PHP Transformer, PHP Transformer Split, Solved Site Promotion Gate all pass), so there is nothing left to repair and this PR now reverts expectations that are correct again.

My error was building on a trunk tip that moved while I was working. Closing.

*AI assistance: diagnosed and closed with Claude Sonnet 4.5 via opencode.*

PR #2069, opened 2026-09-21
fix(blocks-engine): preserve per-side button borders

chubes4:
## Summary

- Rebuild invalid blocks from authoritative comment attributes before serialization.
- Preserve native `core/button` per-side border styles in the saved anchor HTML.
- Add deterministic coverage using Gutenberg `serialize()`, `parse()`, and `validateBlock()` on owner-shaped markup.

## Validation

- `pnpm exec vitest run src/wp/__tests__/button-save-validity.test.ts src/wp/__tests__/canonicalize-inventory.test.ts`
- `pnpm run typecheck`
- `pnpm run build`
- `pnpm test:bundle` runs the new button regression successfully; the existing `core/file` fixture expectation fails because the bundled runtime adds `role="none"` to the sourced RichText marker.

AI contribution: GPT-5.6 Luna, operating directly through OpenCode with operator authorization, implemented and tested this change. Finalization occurred outside Homeboy after control-plane failure.


chubes4:
CI status: the scoped button regression passes, and the fidelity ratchet passes. The required aggregate test job fails only in the pre-existing `src/wp/__tests__/file-save-validity.test.ts` expectation: bundled WordPress adds `role="none"` to a sourced RichText marker. This is unrelated to the canonicalizer/button change and was reproduced locally in both source and bundle test runs. No unrelated compatibility change was made.

chubes4:
Closing as superseded by the production-path PHP fix already merged in #2063 (b5bac7d2). Review confirmed this import executes the PHP transformer, not the TypeScript canonicalizer changed here. The TypeScript repair is not required for the reported import and should not be merged. We will verify the refreshed paired importer against the actual Gutenberg editor. AI assistance: GPT-6 Astra via OpenCode reviewed the owning path and superseding PR; GPT-5.6 Luna via OpenCode supplied the follow-up reproduction.

PR #2097, opened 2026-09-22
fix(php-transformer): preserve responsive typography in artifacts

chubes4:
## Summary

- Fix the artifact compiler path so conditional responsive font sizes are not frozen into inline `fontSize` attributes.
- Keep the deterministic responsive typography carrier class and generated support CSS.
- Add an artifact-path Gutenberg validity regression.

## Verification

- `composer test:canonical --no-interaction`
- `composer test:parity --no-interaction`
- Chromium source/artifact geometry parity at 390/768/1440px: 36/48/60px

## Context

Follow-up to #2090 after the Global Nexus r2 artifact still emitted a frozen `style="font-size:3.75rem"` on the H1.


chubes4:
Superseded by #2098, which is based on current trunk and contains only the artifact-path follow-up commit.

PR #2101, opened 2026-09-22
fix(php-transformer): preserve wrapper geometry

chubes4:
## Summary
- make validity-only synthetic paragraph carriers layout-transparent so source floating links retain their authored boxes
- carry source image classes through `core/media-text` onto its generated image
- add full-artifact regression coverage for footer logo and floating controls

## Verification
- `composer test`
- `php tests/unit/artifact-author-stylesheet-projection.php`
- `php tests/unit/engine-support-css.php`
- `php tests/unit/media-text-block-factory.php`
- `php -l` on all changed PHP source files

The Global Nexus R3 artifact evidence showed the footer logo expanding to `182px x 160px` and synthetic floating-link paragraph carriers establishing extra boxes. The generated artifact now preserves the source image classes and emits the transparency compatibility rule.

Execution provenance: GPT-5.6 Luna via OpenCode, authorized direct recovery outside Homeboy after interruption. Contribution: diagnosis, implementation, regression coverage, and verification.


chubes4:
Native Gutenberg acceptance fails on exact head 49bb56d4c7ba7b3ba7318939da447ba39cf1b2c6: core/media-text save emits no authored class on its inner img, but the candidate stores class="h-9 w-auto object-contain" after stripping the transport attribute from the comment. Browser validation therefore marks core/media-text invalid before and after reload. PHP structural validity is not sufficient for this change. Marked draft pending correction through supported save shape and CSS/selector projection (or appropriate native structure), plus actual browser geometry and editor proof. GPT-6 Astra via OpenCode independently reproduced this in an isolated WordPress fixture; GPT-5.6 Luna direct recovery will address it.

chubes4:
Finalized in commit `0bebd663564778cf8333ee10572c2c6f760a9c5f`. Root cause was authored media classes being serialized onto rigid `core/media-text` inner nodes. The fix keeps the native save shape, carries source classes on the valid outer class carrier, and projects authored image declarations onto `.wp-block-media-text__media > img`. Focused tests, full `composer test`, Gutenberg native save/reload validation, and Playwright geometry checks at 390/768/1440 pass. All exact-head CI checks pass. This authorized recovery/finalization was performed outside Homeboy by GPT-5.6 Luna via OpenCode.

PR #2113, opened 2026-09-22
docs(php-transformer): correct the capture numbers in the media-budget test

gmovr:
Comment-only follow-up to #2107, which merged before the end-to-end verification finished.

Two claims in that test's header comment do not survive measurement:

- **madalenatavares.net** — the request bundle projects **9.9 MiB of text across 48 files and 336.5 MiB of media across 649 files**, not "4.9 MiB text, 335 MiB". The 4.9/335 figures came from a raw `du` of the capture's `website/` directory; what reaches the compiler is the projected bundle.
- **ghestilistes.com** — the two source videos cited for the per-file case (`hd-lb-ultra-stepvideo-hochformat-2411112_1.mp4`, 49 MB, and `EXTENSIONES_INVISIBLES_QUIKKIES.mp4`, 12 MB) live under `resources/external/`, **outside** the captured `website/` tree, and never enter a request bundle. Across every retained capture, no bundle carries a single media file over 10 MiB. That site's actual refusal is `asset-evidence.json` at 10,665,277 bytes — a declared DLA report, textual, genuinely hydrated, so the source-read budget applies to it correctly.

So the per-file media budget is the same bug on the other budget — nothing makes 10 MiB the right ceiling for bytes the compiler does not read — rather than a separately observed failure, and the comment now says that. The assertions and the fixture are unchanged.

Verification of the aggregate case is unaffected: with #2107 and Automattic/static-site-importer#1789, `https://www.madalenatavares.net/` imports end to end (receipt `status: completed`, 6 steps, 22 routes, site created) where it previously never started.

Follow-up to #2107.


chubes4:
Ha ha sorry i merged too fast

PR #2116, opened 2026-09-22
fix(php-transformer): keep a linked image's fill geometry past its link

gmovr:
## Symptom

Source: `https://www.cahoothouse.com/` (run `r41-gate-cahoothouse` / the `vbc` re-import). Reported as "images are not always the right size vs the source".

The home page's "The Cahoot Advantage" cards (Global Partner Venues / Experiences / Off-Sites) render a short photo with a gap below it instead of the photo filling the card's media box:

| | source | import (trunk) |
|---|---|---|
| image rendered | 451x315 | **430x242** |
| image computed height | 300px | 241.797px (its intrinsic ratio at that width) |
| card section height | 624 | **710** |
| grid row height | 395 | 425 |

The same source markup without the link renders correctly, which is the whole tell.

## Root cause

The source card is

```html
<a class="discover-card" href="/experiences/enlisting">
  <div class="image-wrapper"><img class="full-image curated-network" ...></div>
  <div class="card-content-wrap">...</div>
</a>
```

with `.image-wrapper{position:relative;width:100%;min-height:300px;overflow:hidden}` and `.full-image{object-fit:cover;width:100%;height:100%;position:absolute;inset:0%}`.

The compiled block is

```html
<figure class="wp-block-image is-resized full-image curated-network be-inline-geometry-… ">
  <a href="/experiences/enlisting"><img style="object-fit:cover;width:100%;height:100%"></a>
</figure>
```

The figure is right: the projected `.full-image` rule makes it `position:absolute`, and `be-inline-geometry-…{height:100% !important}` gives it the wrapper's 300px. But `core/image` serializes a linked image as `<figure><a></a></figure>`, and that anchor — which the source never had — computes `display:inline-block; position:static; height:auto` (430x242). The image's `height:100%` resolves against *it*, not against the figure, so the image falls back to its intrinsic 2400x1350 ratio.

Two projections stop at the figure and both have to reach through that generated link:

- `php-transformer/src/HtmlToBlocks/Style/StyleResolver.php:1446` `injectedFigureHeightClassName()` — carries `height:100%` onto the injected figure "so authored percentage sizing keeps resolving against the original container box". That is exactly the right idea one box too short: the link is a second injected box in the same chain.
- `php-transformer/src/HtmlToBlocks/Style/AuthorStylesheetProjector.php:1902` `projectImageSelector()` — bridges an author rule that sized the source `` onto the generated one as `<selector>.wp-block-image > img`. A linked image nests the `` one level deeper, so the bridge misses it entirely: the generated image keeps neither the box nor the `object-fit` / `object-position` / `border-radius` the source rule gave it. (cahoothouse's first card is `.full-image.curated-network{object-position:50% 0%}` — the import crops it at `50% 50%`.)

`imageBoxConstraintClassName()` right next door already targets a descendant `img` "so a source anchor still wrapping the image does not matter". These two carriers were the ones that had not caught up.

## Fix

- `injectedFigureHeightClassName()` also restates the fill on the injected link — `.<carrier> > a{display:block !important;width:100% !important;height:100% !important}` — when a source ancestor links the image. The link box is generated and has no source counterpart an author rule could be speaking about, so nothing authored is overridden. An image no ancestor links emits byte-identical CSS to before (no dead rule, same carrier hash).
- `projectImageSelector()` / the root-child and media-text leaf projections emit both native shapes, `… .wp-block-image > img, … .wp-block-image > a > img`, from one `GENERATED_IMAGE_LEAF_PATHS` constant.

### Where the line is drawn

The trigger is the *authored percentage height* on the image (`authorStylesDriveImageHeight()`, unchanged), not any particular fill idiom and not any class name. `position:absolute;inset:0`, a percentage-padding ratio box and a sized flex or grid cell all reach the image through that same percentage, and all three now survive the link (checked: the ratio-box case emits the link rule; the flex-cell-with-only-an-image case never becomes a `core/image` at all — it stays `custom/responsive-media`, verbatim source markup, so it renders correctly and is out of this change's reach).

A source that sizes the image with `aspect-ratio` instead needs no parent height at all; verified in a browser that it renders identically linked and unlinked (400x266.7 either way), so it is deliberately left alone.

The link rule uses `!important` to match the figure rule it accompanies: the engine-support sheet loads *before* the projected author sheet, and this is generated geometry rather than a carried author declaration.

## Tests

`php-transformer/tests/fixtures/parity/html-linked-fill-image-percentage-height-chain.json` — the same card twice in one document, linked and unlinked. It asserts the linked carrier emits the link rule, that the projected bridge covers both leaf paths, and that the unlinked card keeps its exact carrier rule (`not_contains` the link rule) and its exact figure markup. **Fails on trunk** at `assets.0.content`.

Three existing expectations that pinned the single-selector bridge text were updated to the selector list (`tests/contract/run.php`, `tests/unit/artifact-author-stylesheet-projection.php`, `tests/fixtures/parity/html-inline-svg-positioned-css-fill-desktop-mobile.json`).

Full `composer test` + `tests/contract/production-acceptance-matrix.php` pass locally (340 parity fixtures).

## Verification

Related: #2010, #2109, #2111.


gmovr:
End-to-end verification, after merge, for the record — and a follow-up.

I rebuilt the SSI dev package (SSI `edf9d21`, PR Automattic/static-site-importer#1782) and re-imported `https://cahoothouse.com` twice with the Studio CLI, measuring against the live source at 1440 and 390.

**This PR's half, measured on the home page at 1440:**

| | source | before (BE `d0112807`) | after |
|---|---|---|---|
| Advantage card 1 image | 429.88 x 300.00 | **429.88 x 241.80** | 429.88 x 300.00 |
| Advantage card 2 image | 429.88 x 300.00 | **429.88 x 309.84** | 429.88 x 300.00 |
| Advantage card 3 image | 429.88 x 300.00 | **429.88 x 196.00** | 429.88 x 300.00 |
| injected anchor | n/a (source has none) | **429.88 x 241.80, `inline-block`** | 429.88 x 300.00, `block` |
| hero (guard for #2111) | 1440 x 990 | 1440 x 990 | 1440 x 990 |

Same at 390: all three images go from 206.22 / 264.27 / 167.16 to the source's 244.41.

**Follow-up:** #2121. The bridge declarations this PR routed through the link are `object-fit:inherit`, `object-position:inherit` and `border-radius:inherit`, and none of those is an inherited property — so `inherit` reads the generated anchor, which carries none of them. The first Advantage card (`.full-image.curated-network{object-position:50% 0%}`) still imported cropped at `50% 50%` until the anchor relays them. #2121 adds that relay and extends this PR's fixture.

Separately, #2117 covers the other mechanism behind the same report (an `aspect-ratio` ancestor not counting as a definite block size, which collapsed the Lounge / Work-desks cards).

PR #2118, opened 2026-09-22
fix(php-transformer): raise the artifact compiler's per-file ceiling for opaque media references

chubes4:
## What

The artifact compiler's per-file byte ceiling was a flat 10 MiB hard
cap (`ArtifactNormalizer::MAX_FILE_BYTES`) that a caller could not
raise via `compiler_limits.max_file_bytes` — `min()` clamped any
requested value back down to it. A single media file above 10 MiB
could abort an entire site import even when the 320 MiB aggregate
budget (`MAX_TOTAL_BYTES`) had hundreds of MiB of unused headroom.

Real failure from a Studio CLI import (details redacted; source was
private):

```
InvalidArgumentException: A payload reference exceeds the compiler per-file byte limit.
phase: prepare_shared
payloads_retained: 193
```

The bundle held two videos of 13.6 MB and 11.5 MB against ~57 MB of
total media and a 320 MiB aggregate ceiling.

## Investigation

Per the task, I verified rather than assumed how a `payload_reference`
is actually consumed downstream before touching any constant:

- `StagedTransport::prepareReferencedStage()` only calls
  `PayloadReader::read()` (full in-memory buffering) for a reference
  the stage needs to hydrate for parsing, gated by
  `isReferenceBackedBinary()` — which is `false` (i.e. *does* hydrate)
  for text-ish kinds (html/css/js/json/md/svg/text-\*) and `true`
  (stays a reference, never read) for everything else.
- `materializePlanArtifact()` (used during `compose()`) applies the
  same `isReferenceBackedBinary()` gate and skips hydration for opaque
  binaries.
- `WordPressSitePlan`'s asset/write records carry the
  `payload_reference` (schema/id/bytes/sha256) straight through to a
  `referenceWrite()` write record with `payload.encoding = 'reference'`
  — the actual bytes are never read by the compiler; a consumer
  resolves `id` → bytes at materialization time.

**Conclusion: for an opaque binary payload reference (image/audio/
video/font/pdf, not text or SVG), the compiler never buffers it into
memory at any stage.** The "reject unbounded reference declarations
before a reader can allocate" rationale documented on
`assertReferenceLimits()` protects real allocation cost for a
reference the compiler *will* hydrate and parse — it protects nothing
for one it only ever carries by id. For text/parse-oriented
references, the allocation risk is real and the strict ceiling is
kept.

## Fix

- Added `ArtifactNormalizer::MAX_REFERENCE_FILE_BYTES` (= the existing
  `MAX_TOTAL_BYTES`, 320 MiB) — a second, still-bounded ceiling for a
  payload the compiler transports by reference rather than parses.
- `StagedTransport::assertReferenceLimits()` and
  `ArtifactNormalizer`'s per-file oversized check now pick the ceiling
  by payload role (does this file still carry a `payload_reference` /
  is it `isReferenceBackedBinary()`?), not by extension allow-list. A
  caller's `compiler_limits.max_file_bytes` request is honored against
  whichever ceiling applies, instead of being unconditionally clamped
  to the old 10 MiB hard cap.
- The aggregate ceiling (`MAX_TOTAL_BYTES`, and the running `$total`
  check) is untouched and still enforced exactly — this only widens
  the per-file ceiling for opaque payloads.
- Fixed a related silent-drop: `ArtifactNormalizer`'s
  `artifact_file_too_large` path previously always compared against
  the flat parse-oriented ceiling, so it could silently drop an
  admissible media file that the reference-aware `assertReferenceLimits`
  path would have accepted. It now uses the same role-aware ceiling,
  and the surfaced diagnostic reports the ceiling actually applied.
- Fixed a double-clamp: staged preparation re-normalizes an
  already-normalized envelope, feeding the *previous* `limits()`
  output back in as the next call's `compiler_limits`. Since that
  output only preserved the already-narrowed `max_file_bytes`, the
  second pass re-derived a collapsed reference ceiling from it. `limits()`
  now also accepts an explicit `max_reference_file_bytes`, so its own
  output round-trips correctly as input.

This is the last of three caps of the same shape:
- Data Liberation Agent PR #326 — flat per-resource capture cap;
  now scales with the aggregate resource budget.
- Static Site Importer PR #1788 — flat 10 MiB `max_file_bytes` in the
  CLI request-bundle contract; now scales the per-file cap with the
  aggregate budget.

SSI's `tests/smoke-direct-artifact-import.php` already carries a
contract test for the invariant this bug broke: an intake that admits
a file must not hand it to a compiler that refuses it. This PR is what
lets that upstream fix actually take effect.

## Tests

Added coverage in `tests/contract/staged-artifact-compilation.php` and
`tests/unit/artifact-normalizer-source-budget.php` proving:

- a payload reference above the old 10 MiB ceiling is accepted when
  the caller negotiates within the aggregate-scaled bound
- the default (unnegotiated) ceiling still rejects it
- the aggregate ceiling still rejects and still throws, even when two
  references individually fit the raised per-file bound
- a caller-requested limit narrower than the payload is still
  enforced (the negotiated value gates admission, not an unconditional
  maximum)
- a parse-oriented reference (HTML) keeps the strict ceiling even
  when a caller has raised `max_file_bytes` for media
- an oversized payload-referenced file is rejected with a surfaced
  diagnostic naming the ceiling actually applied (both at the
  normalizer level and in a whole-artifact compile's final summary),
  not silently dropped without one

All fixture bytes/hashes in tests are synthetic (`clip.mp4`, etc.); no
real source data.

## CI

- `composer test:unit` — pass
- `composer test:canonical` — pass
- `composer test:parity` — pass (339 fixtures)
- `composer test:wordpress-integration` — passes locally as "skipped:
  WP_TESTS_DIR is unavailable" (no local WP test env); watching CI for
  the real run
- `composer static-parity` intentionally not run locally per task
  instructions (heavy, local-only, not a CI gate)

## AI disclosure

Model: `claude-sonnet-5`. Coding tool: `opencode`. AI wrote the patch
and tests under human orchestration.


chubes4:
Closing as redundant: #2107 ('budget reference-backed media apart from parsed source') landed on trunk while this was in flight and implements the same fix with the same constants — a separate media budget for reference-backed binaries the compiler never parses. Verified trunk accepts the payload sizes that motivated this. No further change needed.

